PR #3057, opened 2026-07-05
fix(qa): wizard ingredient data loss, cookbook create validation, modal a11y labels

adamtasteslikegood:
## Summary

Three fixes from a `/qa` run against production (tasteslikegood.org, 2026-07-04). Full report: `.gstack/qa-reports/qa-report-tasteslikegood-org-2026-07-04.md` (on the QA worktree).

- **ISSUE-002 (High, data loss):** the Write Recipe wizard silently discarded a typed-but-uncommitted ingredient row when clicking Next (and instruction row on Save) unless the green **+** / **Add** button was pressed first. During QA this produced a recipe published publicly with an empty Ingredients section. `nextManualStep()` / `saveManualRecipe()` now auto-commit pending rows; both add methods already no-op on empty input, so committed behavior is unchanged.
- **ISSUE-004 (Medium):** New Cookbook modal: Create with an empty name silently did nothing. The button is now disabled until a name is entered, matching the existing Save Recipe pattern.
- **ISSUE-008 (Medium, a11y):** five icon-only buttons (auth modal close, wizard close, add ingredient, remove ingredient, remove step) had no accessible name. Added aria-labels.

## Verification

- `npm run lint`, `npm run type-check`, `npm test` (51/51), `npm run build` all pass.
- Verified in a local `ng serve` session with headless browser: typed ingredient now survives Next without + (screenshot evidence in QA report), Create correctly disabled/enabled, close buttons findable by their new aria-labels.

## Not in this PR

- ISSUE-001 (public pages unstyled) resolved itself mid-run — v0.3.2 deployed with ebedfd7.
- ISSUE-003: 5 of 28 public browse recipes 404 on `/api/recipes/<id>/image` — backend data issue, needs separate investigation.
- Regression tests for the wizard: component uses `inject()` DI and the vitest setup is node-env without Angular TestBed infra; deferred rather than bootstrapping a new test harness in a QA PR.

---


adamtasteslikegood (commented):
Automated review (Rovo Dev was out of credits per the PR description, so I ran one manually — 5 finder angles + verification against the checked-out branch).

The a11y labels and the New Cookbook disable-guard both check out cleanly against the existing signal/state logic — no issues there.

**One real correctness issue, worth fixing before merge:**

`src/app.component.ts:365-372` — `nextManualStep()`'s new auto-commit, and `saveManualRecipe()`'s at `:402-406`, commit whatever is sitting in `newIngredient()`/`newInstruction()`. But `openManualEntryModal()` (`:345-359`) resets `manualStep`, `manualRecipe`, `manualIngredients`, and `manualInstructions` — it never resets `newIngredient` or `newInstruction`. Neither does `closeManualEntryModal()` (`:361-363`).

Concretely: a user opens the wizard, types an ingredient name on step 2 (e.g. "Old Onion"), then closes the modal via the X instead of clicking `+`/Next. `newIngredient` still holds that value. Next time they open the modal for a *different* recipe, the ingredient list/step reset to empty as expected, but the step-2 input is silently pre-filled with "Old Onion" from the abandoned session. Before this PR, that stale value only leaked in if the user noticed the pre-fill and deliberately clicked `+`. After this PR, merely clicking "Next" (or "Save Recipe", which also unconditionally calls `addManualIngredient()`/`addManualInstruction()`) auto-commits it — silently attaching a leftover ingredient/instruction from an unrelated prior session to the new recipe. Same mechanism for `newInstruction()`.

This is the same class of bug the PR sets out to fix (ISSUE-002, silent data ending up in the wrong place) — just introduced at a different seam. Suggest resetting `newIngredient`/`newInstruction` in `openManualEntryModal()` (and/or `closeManualEntryModal()`) alongside the other manual-entry state.

**Minor, non-blocking:** `saveManualRecipe()`'s own `addManualIngredient()` call (`:405`) is effectively a no-op given the current step gating — the only way to reach step 3 is via `nextManualStep()`'s 2→3 transition, which already flushes `newIngredient`. Not wrong, just duplicate defensive logic wired independently at two call sites keyed to hardcoded step numbers (`manualStep() === 2` / step 3 rendering) rather than a shared "commit pending row for the step being left" helper — worth consolidating if a 4th wizard step is ever added.

---

adamtasteslikegood:
@claude[agent] review against the 3.2 main deployment: since the qa ran while 3.2 was shipping make sure the PR does not cause any regressions or otherwise undo anything that shipped in 3.2

PR #3067, opened 2026-07-08
docs(pm): document CONFLUENCE_SESSION_LOGS_PARENT_ID alias for session-log parent

adamtasteslikegood:
## Summary

Follow-up to #3065, which made `pm_daemon.py` honor the documented `ATLASSIAN_CONFLUENCE_SESSION_LOG_PARENT_PAGE_ID` env var while keeping the original `CONFLUENCE_SESSION_LOGS_PARENT_ID` as a fallback.

`docs/PM_TOOLING.md` still lists only the prefixed name, so an `.env` configured with the unprefixed alias (as existing setups are) reads as misconfigured when audited against the docs. This documents both accepted names and their precedence (prefixed wins if both are set).

Docs-only change — no code touched.

## Context

Surfaced while finishing the Agent Session Logs Confluence template: the session-log parent page (34635777) is wired via `CONFLUENCE_SESSION_LOGS_PARENT_ID` in the local `.env`, which works but wasn't documented.

---


adamtasteslikegood (commented):
## Overview
Docs-only follow-up to #3065. Documents that `pm_daemon.py` accepts `CONFLUENCE_SESSION_LOGS_PARENT_ID` as a fallback alias for `ATLASSIAN_CONFLUENCE_SESSION_LOG_PARENT_PAGE_ID`, with the prefixed name winning if both are set.

## Verification
Checked the claim against the code:

```
scripts/pm/pm_daemon.py:327
SESSION_LOGS_PARENT_ID = os.environ.get('ATLASSIAN_CONFLUENCE_SESSION_LOG_PARENT_PAGE_ID') or os.environ.get('CONFLUENCE_SESSION_LOGS_PARENT_ID', PARENT_PAGE_ID)
```

Confirms exactly what the PR describes: prefixed var wins if set, unprefixed alias is the fallback, then `PARENT_PAGE_ID` as the default. No code changes, so no behavior risk.

## Minor (non-blocking) note
`scripts/pm/publish_session_log.py:38` only reads `ATLASSIAN_CONFLUENCE_SESSION_LOG_PARENT_PAGE_ID` — it doesn't honor the `CONFLUENCE_SESSION_LOGS_PARENT_ID` alias that `pm_daemon.py` supports. That script isn't touched by this PR and isn't described by the section being edited (`## 2. Session Logging (log_agent_session)`, which is specifically the daemon-driven path), so it's out of scope here — just flagging in case a future doc pass wants to call out the inconsistency or align the two scripts.

## Verdict
Accurate, low-risk, correctly scoped. No blocking issues.

(Note: could not formally APPROVE — GitHub disallows self-approval since this token is tied to the PR author's account.)

---

adamtasteslikegood:
@copilot 

Since this PR already targets cloud.md, after reading the file I noticed that it mentions the system health check for Claude sessions and instructs Claude to do a workaround. 

This is no longer an issue since the system health check is now a first-class MCP connector, available the same as any other sanctioned connectors. Please make these changes unless they're already outdated, in which case you can ignore this. 

Thank you.

PR #3093, opened 2026-07-11
fix(ci): recompile issue-arborist workflow

adamtasteslikegood:
## Summary
- regenerate `issue-arborist.lock.yml` from the updated markdown source so the frontmatter hash matches
- retain the built-in `GITHUB_TOKEN` fallback and document why explicit lockdown is disabled
- add the compiler-generated Agentic Maintenance workflow and its pinned `setup-cli` action

## Root cause
PR #3091 changed frontmatter but hand-edited the generated lock file. Every follow-up dispatch stopped in `Check workflow lock file` because the lock recorded hash `87c6f7…` while the source recomputed to `7cfe70…`.

## Validation
- `gh aw compile --strict --validate`
- `gh aw compile --strict --validate --no-emit`
- `git diff --check`

Supersedes the generated-file portion of #3091 and addresses review comment https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/pull/3091#discussion_r3564199579.

---


adamtasteslikegood:
## Review

**What this does:** Recompiles `issue-arborist.lock.yml` from the current markdown source so its `frontmatter_hash` matches (root cause: #3091 hand-edited the lock file instead of running `gh aw compile`, leaving a stale hash that tripped `Check workflow lock file` on every dispatch). Confirmed fixed — the `activation` check run on this PR now passes, where it was previously failing.

**Correctness:** ✅ The only functional delta in `issue-arborist.lock.yml` is removing the leftover `GITHUB_LOCKDOWN_MODE: "1"` env var, consistent with `lockdown: false` set in #3091. `GITHUB_READ_ONLY: "1"` and `toolsets: [issues]` are both still in place, so the MCP server the agent talks to during the run stays read-only regardless. Nothing here re-opens the write access #3091 was trying to close off.

**Worth a second look, not blocking:** the recompile also checked in a new `.github/workflows/agentics-maintenance.yml` (614 lines) — a `gh-aw` compiler side effect, not something this PR's description set out to add. It's scheduled every 6h and its `close-expired-entities` / `cleanup-cache-memory` jobs run unattended with `issues: write`, `pull-requests: write`, `discussions: write`, `actions: write` (the `workflow_dispatch` operations like `disable`/`enable`/`upgrade` are gated behind `check_team_member.cjs`, but the scheduled jobs aren't). Worth confirming this is the automation surface you want running on a cron rather than something to prune before merge — easy to miss since it rode in on a one-line hash fix.

Everything else (the new `setup-cli` pin in `actions-lock.json`, the comment added to `issue-arborist.md`) is exactly what it says on the tin.

---

adamtasteslikegood:
Confirmed, and keeping it. The source declares `safe-outputs.create-issue.expires: 2d`, so a full `gh aw compile --strict --validate` intentionally generates `agentics-maintenance.yml` to enforce expiry and perform gh-aw cache cleanup. The file is byte-for-byte identical to the maintenance workflow produced by Adam's earlier local compile, and its write permissions are scoped to the generated maintenance jobs rather than the Issue Arborist agent job.\n\n_Replied by Copilot on Adam's behalf_

PR #3128, opened 2026-07-15
feat(ci): advisory independent Claude review workflow (cheaper model)

adamtasteslikegood:
Split out of #3125 so that PR stays docs-only. This PR is the executable piece: `.github/workflows/claude-review.yml`. Full design/rationale is in that PR's `docs/ci/refresh/SPEC-03-claude-independent-review.md`.

## What

A **second, independent** PR review that runs in its own CI session on a **different, cheaper model** than the authoring session — author works in Opus 4.8 / Fable 5, the reviewer uses an **Opus 4.x < 4.8** picked deterministically (a `pick` step reads `vars.CLAUDE_REVIEW_MODEL`, defaults to `claude-opus-4-7`, and falls back off the authoring tier if mis-set). It reasons about the diff to choose the review skill (`/code-review`, plus `/security-review` for security-sensitive changes) and reasoning effort, applies fixes, and leaves a comment per fix via `/code-review <effort> --comment --fix` (`anthropics/claude-code-action@v1`).

## Safety / cost

- **Advisory + fail-open** (`continue-on-error`), same-repo + non-bot + non-draft guarded, **never a required check**.
- **Cost-bounded:** once per PR (`opened`/`ready_for_review`) + `claude-review` label + `workflow_dispatch` — not every push; concurrency-cancel; `--max-turns 15`; always the cheaper model.

## Prereqs (maintainer, one-time — inert until done)

1. Add secret **`ANTHROPIC_API_KEY`** (none exists today; repo uses Junie/Gemini/Copilot).
2. Set variable **`CLAUDE_REVIEW_MODEL`** to the confirmed Opus 4.x<8 id — default `claude-opus-4-7` is a **placeholder; verify at console.anthropic.com**.
3. Create the **`claude-review`** label.

⚠️ Verify `/code-review --comment --fix` posts + pushes as intended on a throwaway PR before enabling broadly.

`actionlint` clean.

https://claude.ai/code/session_01SMoNgWobt7jz7YcPjuUrDo

---

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -6,10 +6,22 @@
 # skill + reasoning effort, applies fixes, and leaves a comment per fix.
 #
 # ADVISORY ONLY — never a required status check (see docs/ci/refresh/SPEC-02 §3 and
-# SPEC-03). `continue-on-error: true` so a flaky/absent key never blocks a merge.
+# SPEC-03). `continue-on-error: true` so a flaky/absent key never blocks a merge,
+# and the review step is SKIPPED entirely when no Claude credential is configured.
+#
+# AUTH: prefers a Claude Pro/Max subscription OAuth token, falls back to an API key.
+#   * CLAUDE_CODE_OAUTH_TOKEN (primary) — long-lived token from `claude setup-token`;
+#     draws on your subscription's usage/rate limits.
+#   * ANTHROPIC_API_KEY (fallback) — metered API billing; used when the OAuth secret
+#     is absent. If BOTH are set, the action prefers the OAuth token.
+# `id-token: write` (below) is required so the action can mint the official Claude
+# GitHub App token via OIDC to post comments / push fixes — it is GitHub-side auth,
+# independent of which Claude credential you use.
 #
 # PREREQUISITES (one-time, by a maintainer):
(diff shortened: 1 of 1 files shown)


adamtasteslikegood:
Pushed `d447d2a` to fix the failing run and wire up subscription auth:

- **`id-token: write`** added to `permissions:`. The failure (`Could not fetch an OIDC token`) was GitHub-side auth, not Anthropic: `claude-code-action` mints the official Claude GitHub App token via GitHub OIDC to post comments / push fixes, which requires that permission. This is needed regardless of which Claude credential is used.
- **`CLAUDE_CODE_OAUTH_TOKEN` is now the primary credential** (Pro/Max subscription via `claude setup-token`), with **`ANTHROPIC_API_KEY` kept as fallback**. The action prefers the OAuth token when both are set and falls back to the API key when the OAuth secret is empty.
- **Clean skip when no credential exists**: a pre-flight "Check for a Claude credential" step now skips the review step entirely (instead of running and erroring) when neither secret is set — so forks/Dependabot and unconfigured repos stay green. `continue-on-error: true` remains as a second safety net.

Note: subscription auth draws on your Pro/Max usage/rate limits (shared with interactive Claude Code sessions), whereas the API key isolates CI cost — trade-off called out in the workflow header comment.

PR #3185, opened 2026-07-19
feat(seo): SSR crawlable links on home shell (TAS-2896)- [KAN-114]

adamtasteslikegood:
## What

Adds server-rendered, crawlable anchors to the Angular home shell (`index.html`, built into `dist/index.html`, served by Express at `server/index.ts`).

The shell previously rendered only `<app-root>`, so a crawler (or any no-JS client) hitting `/` found **no anchors** — the public SSR pages (`/browse`, `/r/<slug>`) were a crawl dead-end. This adds real, followable links to the server HTML.

## How

A single `<noscript>` `<nav>` block after `<app-root>` with:
- `<a href="/browse">` — the public recipe index
- two `<a href="/r/<slug>">` recipe links (real published slugs from the live sitemap)

`<noscript>` keeps the links in the server HTML for crawlers/no-JS clients **without changing the JS app's rendered UI** — honoring the scope fence (anchors only; no redesign, no restyle, no new sections).

## Gate

- [x] `grep -q 'href="/browse"' index.html && grep -qE 'href="/r/' index.html` — passes
- [x] Anchors survive `ng build` — present in `dist/index.html` after `npm run build`
- [x] `npm run format:check`, `npm run lint`, `npm run type-check` — pass

_Note: the pre-existing `server/redirects.test.ts > /favicon.ico` failure reproduces identically on clean `dev` (public/ ships `favicon.svg`, no `favicon.ico`) and is unrelated to this change._

Closes TAS-2896.

---
Rovo Dev code review: <strong>Out of Rovo Dev credits</strong>
You've used all your Rovo Dev credits, so Rovo Dev can't review your pull requests.

PR #3195, opened 2026-07-20
fix(ux): show public-recipe View link without publish rights (KAN-119)

adamtasteslikegood:
## Problem

Adam's iOS field report (2026-07-20, post-v0.4.0): the browser still hides the link to a recipe's public page. Root cause: the My Kitchen **View** link renders under `@if (canPublish() && r.is_public)` (`app.component.html:402`), and `canPublish()` requires a signed-in non-guest (`app.component.ts`). Guests — and iOS in-app-webview visitors who *cannot* sign in (Google `disallowed_useragent`; #3186 gives them a fallback panel, not auth) — never see the path to an already-public recipe. Any iOS auth-state hiccup hides it too.

## Fix

Viewing a public page needs no publish rights, so the link's visibility is now a **pure function of recipe data**, with no auth input:

- `src/utils/public-link.ts` — `isPublicViewable(r) = is_public && slug` (slug guard: never `href="/r/undefined"`). Mirrors the `in-app-browser.ts` util+spec pattern from #3186.
- Template `:402` → `@if (isPublicViewable(r))`. The publish **toggle** stays behind `canPublish()` — unchanged.
- TDD: spec written red-first; 4 tests incl. an arity regression guard (no auth parameter).

## Gates (specs/KAN-119_LOOP_PLAN.md, delivery-loop PLAN-OK)

- [x] `npx vitest run src/utils/public-link.spec.ts` — 4/4 (red before impl)
- [x] Scope: exactly 4 src files + loop doc
- [x] `npm run lint` 0 · `npm run type-check` 0 · `npm run build` 0 (noscript anchors survive in `dist/`)
- [x] Full vitest: 133 passed; only the pre-existing `server/redirects.test.ts` favicon failure (documented in #3185 as reproducing on clean dev; `server/` untouched here)

Closes KAN-119. Post-release proving gate: signed-out iOS visit shows View on a published recipe (ux-backlog #6).

https://claude.ai/code/session_01XBVGVmqmwiyTThD4dUSuXs

---

### DIFF
--- a/specs/KAN-119_LOOP_PLAN.md
+++ b/specs/KAN-119_LOOP_PLAN.md
@@ -1,0 +1,46 @@
+# KAN-119 Fix Loop — iOS: public-recipe View link hidden for guests/webview visitors
+
+_2026-07-20 · Owner: Adam · Executor: agent (this loop) · Sprint-2 early-drive (C8), Adam-directed
+ahead of charter lock. Vocabulary: **`fix/`** branch off `dev` (not "hotfix" — that means an
+emergency branch off `main` here; this rides the normal dev → release train)._
+
+## Defect (from KAN-119, code-confirmed)
+
+The My Kitchen "View" link to `/r/<slug>` renders only under `@if (canPublish() && r.is_public)`
+(`src/app.component.html:402`); `canPublish()` requires a signed-in **non-guest**
+(`src/app.component.ts:968-971`). Guests — and iOS in-app-webview visitors who cannot sign in at
+all — never get a path to an already-public recipe's page. Viewing needs no publish rights.
+
+The auth-coupling also means _any_ iOS auth-state hiccup (slow restore, dropped session) hides the
+link. The fix removes the auth dependency entirely, so it is robust to whichever variant Adam hit.
+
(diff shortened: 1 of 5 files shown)


adamtasteslikegood:
Post-merge audit note (Adam's session): the **Independent Claude review job on this PR failed without posting a review** (check concluded `failure`, no comment) — this code merged with green required checks but no independent review. A compensating adversarial review was run on the merged diff (a66dd52). Results:

- **Confirmed OK:** template reachability for guests (the link's block has no enclosing auth gate), no recursion in the delegate (module import resolves, not the class method), negligible change-detection cost (single detail view, not the kitchen list), no data leak (slug only present alongside server-acknowledged `is_public`; `/r/` serves only public recipes).
- **Real gap found:** `buildSavedRecipeFromPublic` copies `sourceSlug` but neither `is_public` nor `slug`, so this fix never fired for the flagship guest case (a guest-saved copy of a public recipe). **Fixed in follow-up #3197** — `publicSlugOf()` resolves the own published slug first, else `sourceSlug`; template href/title resolve through it.

PR #3216, opened 2026-07-23
feat(seo): canonical recipes Phase 0/1 — data file + CI gate + 7 anchors [KAN-118] (KAN-116)

adamtasteslikegood:
## Summary
- **Phase 0/1** of canonical `r/<recipe>` promotion per `specs/CANONICAL_RECIPES_ROLLOUT.md`
- **`specs/canonical-recipes.json`** — tracked data file with slug, title, rationale, and rubric scores for all 7 Adam-approved recipes
- **`scripts/seo/check_canonical_recipes.sh`** — CI validator: checks JSON validity, 3-7 count bound, bidirectional consistency between JSON and `index.html` `<noscript>` anchors, and optional `--live URL` 200 check
- **`pr-gate.yml`** — new `SEO — canonical recipes` job wired into the Gate aggregate
- **`index.html`** — 5 new `<noscript>` anchors added (7 total canonical recipes)
- All 7 slugs verified live at `www.tasteslikegood.org` (200, content checked by Adam)

## Proving gate
- [x] Candidate file exists (`specs/canonical-recipes.json`)
- [x] CI check green (`scripts/seo/check_canonical_recipes.sh` passes)
- [x] All 7 anchors serve live (`--live` check: all 200)
- [x] Adam approved all slugs (confirmed in conversation 2026-07-23)

## Test plan
- [x] `bash scripts/seo/check_canonical_recipes.sh` — "OK: 7 canonical recipes validated"
- [x] `bash scripts/seo/check_canonical_recipes.sh --live https://www.tasteslikegood.org` — all 7 → 200
- [x] `npm run type-check` — clean
- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean
- [x] `npm test` — 146/146 pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
Rovo Dev code review: <strong>Rovo Dev is reviewing this pull request…</strong>
Refresh the page in a few minutes to see the results.

Issue #3209, opened 2026-07-23
refactor(spa): extract shared recipe methods to utility/mixin [KAN-118]


adamtasteslikegood:
## Context

From code review on #3207 (Angular SPA architecture upgrade).

## Problem

`GeneratorComponent` and `RecipeDetailComponent` share ~12 byte-for-byte identical methods: `scaledIngredients`, `scaledServings`, `formatAmount`, `togglePublic`, `exportRecipe`, `startEditNotes`/`cancelEditNotes`/`saveNotes`, `regenerateImage`, `canPublish`, `isPublicViewable`, `publicSlugOf`, `instructionText`.

Any future change must be applied in both places. The `togglePublic` redundant-save bug (fixed in da445b3) was an example of this risk.

## Suggested fix

Extract to a shared utility or base mixin:
- Pure functions (`formatAmount`, `instructionText`, `isPublicViewable`, `publicSlugOf`) → standalone utility
- Stateful methods (`togglePublic`, `exportRecipe`, notes editing) → shared service or mixin

## Files

- `src/components/generator/generator.component.ts`
- `src/components/recipe-detail/recipe-detail.component.ts`

## References

- PR #3207 review comment (claude[bot])
- Jira: KAN-118

PR #3236, opened 2026-07-24
release: v0.4.2 — canonical recipes, SPA architecture, guard fix [KAN-118]

adamtasteslikegood:
## Summary

Sprint 2 close-out release. Promotes `dev` → `main` for v0.4.2 tag + Cloud Build deploy.

### Highlights

- **Canonical recipes Phase 0/1** (KAN-116): tracked schema file with 7 approved slugs, CI validator gate, `<noscript>` home anchors (#3216)
- **Angular Router + component decomposition** (KAN-118): AppComponent 1095→75 lines, lazy loading, 3+ chunks (#3207)
- **Non-blocking SSR save guard** (KAN-118/C9): `/?save=<slug>` paints instantly, network errors retriable (#3214)
- **Backend security** (KAN-64): untrack agent-skill bundles, bump transitive CVE deps (#229, #230, #232)
- **Dependencies**: Angular group (9 pkgs), dd-trace, rate-limit-redis, linting group, google-auth-library, secret-manager, GH Actions (8), body-parser CVE fix
- **Backend pointer**: `18a303a` → `cb218ab` (Backend dev→main #236)

No schema migration.

## Test plan

- [ ] All CI gates pass (Gate, CodeQL, Dependency Review)
- [ ] `package.json` version = `0.4.2`
- [ ] CHANGELOG has `## [0.4.2]` section
- [ ] Tag `v0.4.2` triggers Cloud Build deploy
- [ ] Post-deploy: `curl -s https://www.tasteslikegood.org/ | grep canonical` shows 7 approved slugs

---
Rovo Dev code review: <strong>Rovo Dev couldn't review this pull request</strong>
Something went wrong while reviewing this pull request.
[Troubleshoot code review errors](https://support.atlassian.com/rovo/docs/troubleshoot-rovo-dev-code-reviews/)


adamtasteslikegood (commented):
## Automated review — v0.4.2 release (dev → main)

Reviewed the full diff (61 commits, 65 files). Note that Codex hit its usage limit and Junie's run failed on this PR, so this is the only completed automated pass so far. Nothing here blocks the release; flagging for your call.

### Release hygiene
- `package.json` → `0.4.2` and `CHANGELOG.md` has a matching `## [0.4.2]` section — satisfies the CI changelog gate.
- **Backend submodule pointer**: gitlink moves `18a303ab` → `cb218abf`. Verified `cb218abf` on `adamtasteslikegood/tasteslikegood.com` — it's Backend PR #236, "promote Backend dev→main for cookbook v0.4.2", merged 13:06 UTC (3 min before this PR opened). Its own description says "No schema migration. No functional code changes" (security dep bumps + docs). **Minor nit**: `CHANGELOG.md` cites the pointer as `18a303a → a1891b7`, but `a1891b7` is only one of the commits *inside* #236, not the actual merge SHA. Cosmetic — worth a one-line fix so the changelog matches what's actually pinned, not a blocker.
- `package-lock.json`'s root version is still `0.4.1` (unchanged in the diff) and `CLAUDE.md` still says "currently v0.4.1" — both stale, cosmetic, safe to fix as a fast-follow.

### The guard fix itself (`src/guards/ssr-entry.guard.ts`)
This is the file the PR title calls out as fixed, so I read it closely against `ssr-entry.service.ts`:
- `handleSave()`/`handleAuth()` are fired without `await`, and the guard returns `createUrlTree(['/kitchen'])` immediately after — so navigation to `/kitchen` happens before the save/auth-handoff promise settles. To be clear, this **isn't silent**: `SsrEntryService.handleSave` does call `toast.show(...)` on the fetch-failure and catch paths, so a failed save still surfaces a toast — just after the user has already landed on `/kitchen`, which can read as mildly disorienting UX rather than a lost save.
- One real gap: `await this.auth.ready` and `this.auth.ensureGuestSession()` run *outside* the service's `try/catch`, so if either throws/rejects, it only hits the guard's `.catch(err => console.error(...))` — no toast in that specific case.
- Separately: `save` and `auth=success` are handled as independent early-return `if` blocks, so a URL carrying both simultaneously only processes `save` and silently drops the auth handoff. Worth checking whether that combination is actually reachable from any real CTA link.
- No test file exists for this guard (`ssr-entry.guard.test.ts`/`.spec.ts` not present anywhere in the diff), so none of the above is regression-protected.

Suggest (non-blocking, could be a fast-follow): `return ssrEntry.handleSave(slug).then(() => router.createUrlTree(['/kitchen'])).catch(() => router.createUrlTree(['/kitchen']))` so the toast lands before/at the same time as navigation, plus a basic spec.

### Other things noticed, lower priority
- `src/components/recipe-detail/recipe-detail.component.ts` and `src/components/header/header.component.ts` both use raw RxJS `.subscribe()` without `takeUntilDestroyed()`, which is a direct deviation from this repo's "Signals API, no RxJS" convention (low leak risk since `header` is a shell singleton, but worth a look).
- `generator.component.ts` `togglePublic()` mutates `recipe.is_public`/`recipe.slug` directly instead of through the signal's `.set()`/`.update()` like its sibling methods, and its rollback-on-failure path isn't awaited.
- Kitchen's Recycle Bin permanent-delete button reportedly has no confirmation step (unlike the adjacent soft-delete flow) — worth a quick manual check before shipping if that's accurate, since it'd be a one-tap irreversible delete.

No secrets, XSS vectors, or unsafe patterns found in the CI workflow changes or the app code otherwise. Dependency Review already came back clean.

---

adamtasteslikegood:
**Re: Atlassian bot inline comment on `recipe-detail.component.ts:73` (`paramMap.subscribe` without `takeUntilDestroyed`)**

Not a real leak — Angular's `ActivatedRoute` is scoped to the routed component and its `paramMap` observable completes when the component is destroyed. The independent Claude review confirmed this: "subscriptions auto-complete on destroy."

Adding `takeUntilDestroyed()` is a valid defensive-coding pattern but would also introduce an RxJS import into a component that currently uses zero RxJS (pure Signals). Deferring to backlog alongside the existing #3209 (shared recipe method extraction).

adamtasteslikegood:
**Re: automated owner review + independent Claude review** — feedback triaged; fixes are up in #3238 (targets `dev`, so this release PR picks them up on merge). Point by point:

**Fixed in #3238:**
- **CHANGELOG pointer SHA** (both reviews): now cites `18a303a → cb218ab` (Backend dev→main promotion tasteslikegood.com#236) instead of the intermediate `a1891b7`.
- **`package-lock.json` root version** synced to `0.4.2`; **CLAUDE.md** "currently v0.4.1" note bumped.
- **Toast-less failure gap in `handleSave`** — confirmed real: `await this.auth.ready` / `ensureGuestSession()` sat outside the try/catch, so a rejection there only hit the guard's `console.error` with no user feedback. Now caught with the generic error toast. Covered by a new service test.
- **No guard test** — added `src/guards/ssr-entry.guard.test.ts` (7 cases: save redirect, repeated-param, rejection logging, auth handoff, save/auth precedence, `#kitchen` fragment, pass-through).

**Not changing, with rationale:**
- **Fire-and-forget save before navigation**: this is the deliberate non-blocking design shipped in #3214 ("`/?save=<slug>` paints instantly") — awaiting `handleSave` before returning the `UrlTree` would reintroduce the blocking entry paint that PR existed to remove. The toast intentionally lands just after `/kitchen` renders. The new guard test pins this behavior.
- **`save` + `auth=success` combined**: not reachable from any real flow — the OAuth callback redirects to a bare `{frontend_url}?auth=success` (`Backend/blueprints/auth_api_bp.py:303`) and the SSR CTA emits only `/?save=<slug>`. Even if hand-crafted, `handleSave` awaits `this.auth.ready`, which is the entirety of what `handleAuth` does, so nothing is lost beyond the redirect target. The precedence is now documented by a test.
- **`takeUntilDestroyed` on header/recipe-detail subscriptions**: as replied earlier on the Atlassian bot thread — `ActivatedRoute.paramMap` completes on component destroy and header is a shell singleton; deferring to backlog rather than adding an RxJS import to pure-Signals components.
- **`togglePublic` direct mutation / un-awaited rollback**: confirmed accurate (`generator.component.ts:187-209`), but it's pre-existing behavior carried through the decomposition, not a regression in this release — deferring to the backlog alongside #3209 (shared recipe method extraction) rather than changing publish semantics inside a release PR.
- **Recycle-bin per-item permanent delete**: verified accurate — the per-item button (`kitchen.component.html:222`) calls `permanentlyDeleteRecipe()` directly with no confirm, while the bulk "empty bin" flow does have a confirmation modal. Items there are already behind an explicit soft-delete step, so one-tap may be intended — flagging as a product call for Adam; happy to add a confirm reusing the existing modal as a fast-follow if wanted.

PR #3244, opened 2026-07-24
fix(ux): saved-copy view/publish state cluster — differentiated View link, publish confirm, isSaved fix, guest publish affordance [KAN-137]

### LINKED ISSUE #3210: fix(spa): viewRecipe sets isSaved=true for non-owned deep-link recipes [KAN-118]
## Context

From code review on #3207 (Angular SPA architecture upgrade).

## Problem

`RecipeStateService.viewRecipe(r)` unconditionally sets `isSaved = true`. When `RecipeDetailComponent.fetchRecipeFromApi(id)` calls this for a recipe fetched via deep link that isn't in the user's cookbook, `isSaved` is misreported as `true`, which disables the Save button.

## Suggested fix

Add an explicit `saved` parameter to `viewRecipe`:

```ts
viewRecipe(r: Recipe, saved = true) {
  this.currentRecipe.set(r);
  this.isSaved.set(saved);
}
```

And call `viewRecipe(recipe, false)` from the API fetch fallback path.

## Files

- `src/services/recipe-state.service.ts`
- `src/components/recipe-detail/recipe-detail.component.ts`

## References

- PR #3207 review comment (claude[bot])
- Jira: KAN-118

### LINKED ISSUE #3211: fix(ux): add 'Sign in to publish' button to recipe-detail view [KAN-118]
## Context

From code review on #3207 (Angular SPA architecture upgrade).

## Problem

In `generator.component.html`, the `@if (canPublish())` block is paired with an `@else if (isSaved())` that renders a "Sign in to publish" button and opens the auth modal. In `recipe-detail.component.html`, the else branch is missing entirely — a guest opening a saved recipe via `/recipe/:id` sees no publish affordance at all.

## Suggested fix

Mirror the generator's `@else if (isSaved())` block in recipe-detail, injecting ModalService to open the auth modal.

## Files

- `src/components/recipe-detail/recipe-detail.component.html`
- `src/components/recipe-detail/recipe-detail.component.ts`

## References

- PR #3207 review comment (claude[bot])
- Jira: KAN-118, KAN-119 (iOS view-gating)

adamtasteslikegood:
## Summary

Sprint 3 item B, first cluster (Jira **KAN-137**, epic KAN-136). Fixes the cross-author save→publish→view cycle Adam reproduced live, enforcing his invariant: **the public toggle must never read OFF next to an undifferentiated View link.**

### Root causes (systematic-debugging pass)
1. The KAN-119 `sourceSlug` fallback makes a saved copy's View link (which opens the *original's* page) indistinguishable from an own-published View link — so publish-capable users see toggle-off + View-present.
2. Publishing a saved copy silently mints a near-identical second public page (name collision → `-N` slug). Unpublishing keeps the slug but drops `is_public`, falling back to the sourceSlug link — cycling back to state 1.
3. GH #3210: `viewRecipe` unconditionally set `isSaved=true` on cold deep-link fetches.
4. GH #3211: recipe-detail had no publish affordance for guests.

### Changes
- `publicLinkKind()` ('own' | 'source' | null) in `public-link.ts`; both generator and recipe-detail render 'source' View links **muted + italic** with tooltip *"This recipe was saved from a public recipe — opens the original"* (Adam's chosen option: differentiated, not hidden — guests keep their KAN-119 link).
- `togglePublic` (both components): first publish of a `sourceSlug` copy requires a `confirm()` acknowledging the original may still be live (same pattern as kitchen's delete confirm).
- `viewRecipe(r, saved = false)` on the API-fetch path — closes #3210.
- Recipe-detail mirrors generator's *Sign in to publish* button via ModalService — closes #3211.

### Tests
- `public-link.spec.ts`: full `publicLinkKind` matrix incl. the unpublish-leg regression case, and a visibility-agreement property with `isPublicViewable`.
- New `recipe-state.service.test.ts`: saved-default, deep-link-unsaved, clear-reset.
- Existing recipe-detail error-handling tests updated for the ModalService provider.
- Suite: 162/163 (the 1 failure is `server/redirects.test.ts` favicon — an artifact of running from a `.claude/worktrees/` path where Express `send` ignores dot-segments; passes in the main checkout and CI).

### Verification for Adam
Walkthrough: save someone else's public recipe → copy shows muted italic View (source) + toggle OFF (consistent) → toggle ON asks confirmation → after publish, View turns normal (own page) → unpublish → link degrades back to muted source style, toggle OFF. No silent duplicate pages.

Closes #3210. Closes #3211.

---

### DIFF
--- a/src/components/generator/generator.component.html
+++ b/src/components/generator/generator.component.html
@@ -203,11 +203,9 @@
                   [href]="'/r/' + publicSlugOf(r)"
                   target="_blank"
                   rel="noopener noreferrer"
-                  class="inline-flex items-center gap-1 ml-1 text-xs font-semibold transition-colors animate-[fadeIn_0.2s]"
-                  [class.text-stone-500]="linkKind === 'own'"
+                  class="inline-flex items-center gap-1 ml-1 text-xs font-semibold text-stone-500 transition-colors animate-[fadeIn_0.2s]"
                   [class.hover:text-green-600]="linkKind === 'own'"
-                  [class.text-stone-300]="linkKind === 'source'"
-                  [class.hover:text-stone-400]="linkKind === 'source'"
+                  [class.hover:text-stone-700]="linkKind === 'source'"
                   [class.italic]="linkKind === 'source'"
                   [title]="
                     linkKind === 'source'
--- a/src/components/generator/generator.component.html
+++ b/src/components/generator/generator.component.html
@@ -195,13 +195,26 @@
                   Sign in to publish
                 </button>
               }
-              @if (isPublicViewable(r)) {
+              @if (publicLinkKind(r); as linkKind) {
+                <!-- KAN-137: a 'source' link opens the page this copy was
+                     saved FROM — muted so it never reads as this recipe's own
+                     published page while the toggle is off. -->
                 <a
                   [href]="'/r/' + publicSlugOf(r)"
                   target="_blank"
                   rel="noopener noreferrer"
-                  class="inline-flex items-center gap-1 ml-1 text-xs font-semibold text-stone-500 hover:text-green-600 transition-colors animate-[fadeIn_0.2s]"
-                  [title]="'View public page: /r/' + publicSlugOf(r)"
+                  class="inline-flex items-center gap-1 ml-1 text-xs font-semibold transition-colors animate-[fadeIn_0.2s]"
+                  [class.text-stone-500]="linkKind === 'own'"
+                  [class.hover:text-green-600]="linkKind === 'own'"
+                  [class.text-stone-300]="linkKind === 'source'"
+                  [class.hover:text-stone-400]="linkKind === 'source'"
+                  [class.italic]="linkKind === 'source'"
+                  [title]="
+                    linkKind === 'source'
+                      ? 'This recipe was saved from a public recipe — opens the original: /r/' +
+                        publicSlugOf(r)
+                      : 'View public page: /r/' + publicSlugOf(r)
+                  "
                 >
                   View
                   <svg
--- a/src/components/generator/generator.component.ts
+++ b/src/components/generator/generator.component.ts
@@ -6,7 +6,7 @@ import { AuthService } from '../../services/auth.service';
 import { PersistenceService } from '../../services/persistence.service';
 import { RecipeStateService } from '../../services/recipe-state.service';
 import { ModalService } from '../../services/modal.service';
-import { isPublicViewable, publicSlugOf } from '../../utils/public-link';
+import { isPublicViewable, publicLinkKind, publicSlugOf } from '../../utils/public-link';
 import { slugFromTitle } from '../../utils/slug';
 import type { Ingredient, IngredientGroup, InstructionStep, Recipe } from '../../recipe.types';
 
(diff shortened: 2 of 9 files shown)


adamtasteslikegood (commented):
## Review: saved-copy view/publish state cluster (KAN-137)

### Overview
Fixes a three-part invariant violation where the public View link and the publish toggle could disagree for a recipe saved from someone else's published page. Adds `publicLinkKind()` (`'own' | 'source' | null`) to visually differentiate a link that opens the *original* recipe from one that opens the viewer's *own* published page, gates first-publish of a `sourceSlug` copy behind a `confirm()` (avoiding a silent duplicate public page), fixes `RecipeStateService.viewRecipe` unconditionally marking cold API-fetched recipes as saved (#3210), and adds a "Sign in to publish" affordance to recipe-detail for guests (#3211).

### Correctness
- `publicLinkKind()`'s precedence (`is_public && slug` wins over `sourceSlug`) is consistent with the existing `publicSlugOf()`/`isPublicViewable()` logic — confirmed by the added "agrees with publicSlugOf/isPublicViewable" property test.
- The `togglePublic` confirm-guard (`nextState && !recipe.slug && recipe.sourceSlug`) fires only on first-publish of a `sourceSlug` copy; re-publishing after an unpublish (slug already assigned) correctly skips the dialog, matching the described "unpublish leg" behavior.
- `viewRecipe(r, saved = true)` keeps the default for the existing cookbook-navigation call site and only flips to `false` at the one new cold-fetch call site — minimal, backward-compatible.

### Suggestions (non-blocking)
1. **No unit test exercises the new `confirm()` branch** in `togglePublic` (either component). `public-link.spec.ts` / `recipe-state.service.test.ts` cover the pure-function logic well, but nothing asserts the early-return-on-cancel or proceeds-on-accept paths. jsdom's default `window.confirm` returns `undefined` rather than throwing, so `vi.stubGlobal('confirm', vi.fn(() => false))` would be an easy addition given this is a new side-effect gate on a mutating path.
2. **Duplication across `GeneratorComponent`/`RecipeDetailComponent`**: `togglePublic`, `publicLinkKind`, `isPublicViewable`, `publicSlugOf`, and the confirm-guard are now duplicated verbatim in both components. This extends a pre-existing pattern rather than introducing new duplication, and matches the project's surgical-diff convention — just flagging that a third publish-capable component would be a good trigger to extract shared logic.
3. **Guest affordance is gated on `isSaved()`**: since cold API-fetched recipes now default to `isSaved() === false` (the #3210 fix), a guest opening a not-yet-saved recipe via deep link sees neither the toggle nor "Sign in to publish" — only a recipe already in the guest's local cookbook shows it. Likely intentional (nothing to publish until it's saved), but worth a quick sanity check against the original #3211 repro to confirm it's covered.
4. Minor, pre-existing and untouched by this diff: `recipe.slug = this.slugFromTitle(recipe.name)` sets a client-side slug guess before the server round-trip; if the server appends a `-N` disambiguator on collision, the locally-set slug could momentarily mismatch what's persisted.

### Tests / CI
Suite reported 162/163 passing; the one failure is attributed to a known worktree-path artifact and looks safe to disregard. At review time `Frontend — build`, `Dependency Review`, and `Independent Claude review` were still in progress — worth a final glance once those land.

### Security
No concerns — `confirm()` usage matches the existing kitchen-delete pattern, and all new template bindings are standard Angular property bindings (no `innerHTML`/`bypassSecurityTrust`).

**Verdict:** Solid, well-scoped fix with good coverage of the pure logic. Suggestions above are non-blocking.

---

adamtasteslikegood:
Re: independent review — all points dispositioned:

1. **WCAG contrast fix (287e78a): verified and kept.** Checked the diff and re-ran the suite locally on top of it (78/78 src tests, Prettier clean). The italic + tooltip + neutral-hover differentiation preserves the KAN-137 'source vs own' distinction through a non-chromatic cue, which is strictly better than my original stone-300 styling. No changes needed.

2. **Confirm-guard test gap: fixed in 03beeaa.** Four new tests lock the invariant — confirm fires only on first publish of a sourceSlug copy (message includes the source /r/ path), a 'no' leaves `is_public` and persistence untouched, own publishes and unpublishes never prompt.

3. **Duplicated togglePublic/View-link block: agreed, deferred.** This is exactly GH #3209's territory (extract shared recipe methods — already a Sprint-3-parked refactor). Noted there rather than widening this bug-fix PR.

4. **'Sign in to publish' contrast: agreed, deferred.** Pre-existing pattern shared with the generator; piecemeal-fixing it here would desync the two components. Belongs to the broader a11y sweep with the rest of the stone-400 text.

*Replied by Claude on Adam's behalf*

Issue #3147, opened 2026-07-17
Decide: should renaming a recipe (or an explicit action) ever change its public /r/<slug> URL?


adamtasteslikegood:
Product decision surfaced during review of #3138.

## Current behavior (after #3138)

A recipe's slug is **permanent once assigned**:

- The client only derives a slug when the recipe has none (`togglePublic()`: `if (nextState && !recipe.slug)`).
- The server keeps the existing slug on every later save — `_resolve_public_slug` (`Backend/repositories/db_recipe_repository.py`) returns `current_slug` whenever it is salvageable, regardless of title changes.
- Unpublishing does not clear the slug, so unpublish → rename → republish still yields the **old** URL.

So after #3138 removed the editable slug field, there is **no path at all** to change a recipe's public URL. The PR originally described renaming as the way to change it ("rename the recipe to change it"), which is not what the code does — the template comment was corrected in ca3eacc to state the slug is permanent.

## The decision

Permanence is arguably correct for a stable, indexed, shareable URL (no link rot, no SEO churn). But it should be a deliberate choice, not an accident of implementation. Options:

1. **Keep permanence (status quo)** — document it as intended; done, close this issue.
2. **Re-derive on rename** — title changes update the slug (needs a redirect story for the old URL: 301 from old slug, or slug history table, to avoid breaking shared/indexed links).
3. **Explicit owner action** — a "regenerate URL" affordance in My Kitchen, keeping day-to-day renames URL-stable but giving owners an escape hatch (same redirect question as 2; also the natural home for fixing the non-Latin-title dead-end in #3146).

_Opened by Claude on Adam's behalf_

Jira: KAN-105

adamtasteslikegood:
## Recommendation: **No — a slug is permanent once minted.** Renaming never changes it, and no automatic re-slugging exists on any path.

Drafted for Adam's decision, informed by what shipped this week (KAN-137/139/140):

### Proposed policy

1. **Slug = identity, title = display.** Once a recipe has minted `/r/<slug>`, that URL never changes for the life of the row — same model as WordPress/Ghost, where editing the title doesn't touch the published URL.
2. **Canonical recipes: already decided.** `is_canonical` (KAN-139) hard-locks slug + publish state server-side. This issue's question is only about the *non-canonical* remainder.
[...]

PR #3251, opened 2026-07-24
fix(spa): surface publish failures — empty-slug pre-check + sync-failure toast [KAN-104]

### LINKED ISSUE #3146: Publish flow: surface the server 400 when a title derives to an empty slug (non-Latin titles publish with a dead /r/ link)
Follow-up flagged during review of #3138 (see [this inline thread](https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/pull/3138#discussion_r3595305965)) — deliberately not fixed there because the root cause predates that PR and sits outside its scope.

## Problem

`slugFromTitle` returns `""` for titles with no ASCII alphanumerics after NFKD + ASCII-strip — all-emoji, pure-CJK, pure-Cyrillic titles (e.g. `豆腐カレー`, `Борщ`, `🌮🌮🌮`). Plausible titles for a cookbook with international recipes.

The failure is then **silent**:

1. `togglePublic()` (`src/app.component.ts`) optimistically sets `recipe.is_public = true` before the POST.
2. The server correctly rejects with 400 `PUBLIC_SLUG_REQUIRED_ERROR` (`Backend/repositories/db_recipe_repository.py`, `_resolve_public_slug`).
3. `PersistenceService._apiSaveRecipe` only `console.warn`s on non-2xx — it never throws — so the `catch` in `togglePublic()` that reverts `is_public` never fires.
4. The UI shows the recipe as published with a dead `/r/` link until the next full reload re-hydrates from the API.

Aggravating factor: #3138 removed the free-form slug input, which was previously the only workaround (manually typing an ASCII slug). Such recipes can now only be published by renaming the title to include ASCII characters — with no hint to the user that this is why publishing "didn't work".

The parity spec added in ca3eacc pins the two empty-slug cases as expected client behavior, so this issue is purely about error surfacing, not derivation.

## Suggested direction

- Promote non-2xx in `_apiSaveRecipe` to a thrown error (or a returned result) so `togglePublic()`'s existing revert path actually runs — note the method has three other call sites (`saveRecipe`, `restoreRecipe`, `addRecipeToCookbook`) whose swallowing semantics need a decision.
- Revert the optimistic `is_public` flip and tell the user why (the server's fixed `PUBLIC_SLUG_REQUIRED_ERROR` message is already client-safe).
- Optionally: disable/guard the publish toggle client-side when `slugFromTitle(recipe.name) === ''`, as a cheap pre-check.

Jira: KAN-104

adamtasteslikegood:
Closes #3146.

Two of the issue's three suggested directions, adjusted for what already shipped since it was filed:

1. **Client pre-check** — `togglePublic` (both generator and recipe-detail) runs `slugFromTitle(recipe.name)` before flipping anything: a title with no ASCII alphanumerics gets the toast *"This recipe can't be published: its title has no letters or numbers (a-z, 0-9) to build a public link from."* and no request is made. Client/server derivation parity is pinned by the existing slug spec, so the pre-check is authoritative for this failure class.
2. **Failure surfacing** — the optimistic-revert half of the issue was already fixed (KAN-119 era: `saveRecipe` resolves `false`, `togglePublic` reverts) but the revert was **silent**. It now shows "Publishing/Unpublishing failed to sync to the server. Check your connection and try again."

Not done: propagating the server's `PUBLIC_SLUG_REQUIRED_ERROR` body through `_apiSaveRecipe`'s boolean contract — with the pre-check in place that 400 is unreachable for this failure class, and widening the return type would touch three other call sites for no user-visible gain (YAGNI).

**Related finding** (feeds #3147): the UI currently has **no rename affordance for existing recipes** — the manual-entry modal always creates a new recipe (fresh UUID), and generator/detail only edit notes. So the only way a user hits this toast is at manual-create time, and the only 'fix' is re-creating the recipe with a usable title.

## Verification

- New tests: empty-slug pre-check (blocks, toasts, no save) + failed-sync revert toast — src suite **90/90 green**
- lint / format / type-check clean

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixed
+
+- **Publish failures are no longer silent** (KAN-104,
+  [#3146](https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/issues/3146)):
+  titles with no ASCII letters/numbers (all-emoji, pure-CJK/Cyrillic) are
+  pre-checked client-side with an explanatory toast instead of hitting the
+  server's 400 silently, and a publish/unpublish that fails to sync now
+  reverts _and_ tells the user (previously it only reverted).
+
 ### Added
 
 - **Publish state resolves to one DB row** (KAN-139,
--- a/src/components/generator/generator.component.ts
+++ b/src/components/generator/generator.component.ts
@@ -6,6 +6,7 @@ import { AuthService } from '../../services/auth.service';
 import { PersistenceService } from '../../services/persistence.service';
 import { RecipeStateService } from '../../services/recipe-state.service';
 import { ModalService } from '../../services/modal.service';
+import { ToastService } from '../../services/toast.service';
 import {
   isPublicViewable,
   publicLinkKind,
@@ -27,6 +28,7 @@ export class GeneratorComponent {
   readonly authService = inject(AuthService);
   private readonly recipeState = inject(RecipeStateService);
   private readonly modalService = inject(ModalService);
+  private readonly toastService = inject(ToastService);
 
   readonly recipe = this.recipeState.currentRecipe;
   readonly generatedImageUrl = this.recipeState.generatedImageUrl;
@@ -201,6 +203,17 @@ export class GeneratorComponent {
     if (recipe.is_canonical || this.publishTogglePending()) return;
     const nextState = !recipe.is_public;
 
+    // KAN-104 (#3146): a title with no ASCII alphanumerics (all-emoji,
+    // pure-CJK/Cyrillic) derives an empty slug, which the server rejects
+    // with 400. Same derivation as the server (parity is spec-pinned), so
+    // catch it before the round trip and say why instead of failing silently.
+    if (nextState && !recipe.slug && !this.slugFromTitle(recipe.name)) {
+      this.toastService.show(
+        "This recipe can't be published: its title has no letters or numbers (a-z, 0-9) to build a public link from."
+      );
+      return;
+    }
+
     // KAN-137: first publish of a copy saved from a public recipe would mint
     // a near-identical second public page (name collision → -N slug). Make
     // that an informed choice instead of a silent side effect.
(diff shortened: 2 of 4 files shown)


adamtasteslikegood (commented):
## Overview

Two-part fix for silent publish failures (KAN-104, closes #3146):

1. **Client pre-check** — `togglePublic` in both `generator.component.ts` and `recipe-detail.component.ts` now runs `slugFromTitle(recipe.name)` before flipping any state; an empty result (title has no ASCII alphanumerics) shows an explanatory toast and aborts before the request.
2. **Failure surfacing** — the existing optimistic-revert-on-sync-failure path (from the KAN-119 era) now shows a toast instead of silently snapping the toggle back.

## Code quality / correctness

- The pre-check is correctly gated (`nextState && !recipe.slug && !slugFromTitle(...)`) so it only fires on the publish direction with no existing slug, and it runs before the KAN-137 `sourceSlug` confirm-guard and before any mutation of `recipe.is_public` — matches the new tests (no `confirm()` call, `is_public` stays falsy, `saveRecipe` not called).
- Reusing `slugFromTitle` for both the pre-check and the later `recipe.slug = this.slugFromTitle(recipe.name)` assignment correctly keeps client/server derivation parity, per the existing slug spec.
- The catch-block toast reuses `nextState` (captured before mutation) to pick "Publishing" vs "Unpublishing" wording — correct, since by the time the catch runs `recipe.is_public` has already been reverted.
- CHANGELOG entry is clear and correctly scoped under `[Unreleased] → Fixed`.

## Suggestions (non-blocking)

- **Test coverage gap**: the new logic is duplicated verbatim into `generator.component.ts`, but the new tests only live in `recipe-detail.component.test.ts` — there's no `generator.component.test.ts` at all, so the identical empty-slug pre-check and failed-sync toast in the generator path ship untested. Pre-existing gap (the component had zero test coverage before this PR), but since this PR adds new branching logic there, it'd be worth at least mirroring the two new recipe-detail tests for the generator component.
- **Minor**: `slugFromTitle(recipe.name)` is computed twice on the publish path (once in the pre-check guard, once a few lines later when assigning `recipe.slug`). Not a correctness issue, just a redundant call — could hoist into a local if it bugs anyone, but not worth a diff on its own.

## Risks

- None that block merge. The pre-check can't introduce a regression for existing published recipes (guarded on `!recipe.slug`), and the toast-only change to the catch block is additive — it doesn't alter control flow.

Verification claims in the PR body (90/90 src tests, lint/format/type-check clean) look consistent with the diff shown.


---

PR #3265, opened 2026-07-24
fix(spa): publish-state cluster — server-owned slugs, deep-link row normalization, View-link overflow (#3262, #3263, #3264) [KAN-149]

### LINKED ISSUE #3262: fix(spa): View link uses client-predicted slug on publish — wrong page/404 on name collision [KAN-136]
## Adam's post-v0.4.4 field test (2026-07-24)

Sequence: unpublish `vegan-zucchini-poppers` (old row) → 404 ✓ → generate a new "Zucchini Poppers" → publish it → **SPA View link points at `/r/vegan-zucchini-poppers` (not `-2`) and 404s** → republish the old row → base slug 200s again and **both recipes' links point at the same public page** (the OLD recipe's). Unpublishing the new recipe afterwards didn't 404 the base slug (it belongs to the old row).

## Root cause

Legacy client-side slug prediction, present in both components (predates the SPA extraction — carried out of the monolith):

```ts
// generator.component.ts:246, recipe-detail.component.ts:218
if (nextState && !recipe.slug) {
  recipe.slug = this.slugFromTitle(recipe.name);
}
```

The server owns slug minting (`_resolve_public_slug`): on name collision with ANY existing row — published or not, since unpublished rows retain their slug (KAN-139 semantics, table-wide unique) — it mints `-2`. The SPA never merges the server's answer back after the publish sync, so the local blob keeps the predicted base slug and the View link renders it:

- 404 while the base slug's owner is unpublished
- **wrong page** (someone else's/the old recipe's) once the base slug is live
- two local recipes can display the identical predicted link — the server never double-assigns (unique constraint); only the client display collides

A full reload self-heals: `loadFromApi`'s column-over-blob merge (KAN-139) replaces the predicted slug with server truth. The gap is between publish and next reload.

## Fix direction

Stop predicting: leave `recipe.slug` unset on publish and merge the server-returned row (slug included) into local state when the publish sync resolves — the `publishStateSync` infrastructure (KAN-139) already marks the sync lifecycle; the View link can render greyed/pending until the real slug arrives, the same pattern as the toggle.

Note: server-side integrity is unaffected (server ignores client slug on save — KAN-103 #3138); this is a display/state bug only.

### LINKED ISSUE #3263: fix(spa): refresh on /recipe/:id renders a blank page — cold fetch treats API row as recipe blob [KAN-136]
## Adam's report (post-v0.4.4, present since the Angular upgrade #3207)

Browser refresh while on an SPA recipe page (`/recipe/:id`) renders a blank page.

## Root cause (code-verified, needs browser confirm)

Express serves the shell fine (verified: `/recipe/<id>` returns 200 + `<app-root>` + current bundle). The failure is client-side, in the cold deep-link path added with the router (#3207):

`recipe-detail.component.ts` `fetchRecipeFromApi()` does:

```ts
const recipe: Recipe = await resp.json();
this.recipeState.viewRecipe(recipe, false);
```

but `GET /api/recipes/:id` returns the **row shape** (`recipe.to_dict()`: `{id, name, slug, is_public, is_canonical, origin, data: {…blob…}}`), not the Recipe blob. So `r.ingredients`, `r.instructions`, `r.tags` etc. are `undefined` (they live under `r.data`), `r.servings` is undefined (→ NaN math), and the template's `@for` loops over undefined throw at render → blank page.

Why every refresh hits it (not just true cold links): on a cold boot the constructor's `paramMap` handler runs before `authService.currentUser()` has hydrated, so even recipes that ARE in the cookbook miss the `savedRecipes` branch and fall through to the API fetch.

## Fix direction

Normalize the fetch the same way `loadFromApi` does (KAN-139 column-over-blob merge): unwrap `row.data` and overlay the authoritative columns (`slug`, `is_public`, `is_canonical`, `source_slug`, `origin`) before handing it to `viewRecipe`. Extract that merge into a shared util so persistence-service and recipe-detail can't drift — which is exactly the #3209 shared-extraction argument again.

Related: #3262 (slug prediction), #3209.

### LINKED ISSUE #3264: fix(ux): toggling publish ON pushes the View link behind the recipe image until refresh [KAN-136]
## Adam's report (post-v0.4.4, present since the Angular upgrade #3207)

On the generator card, toggling publish ON makes the newly-mounted **View** link render partially/fully behind the recipe image until the page is revisited or refreshed.

## What's known from the template

The View link mounts conditionally (`@if (publicLinkKind(r); as linkKind)`) at the end of the distribution-controls flex cluster (`generator.component.html` ~line 211), with `animate-[fadeIn_0.2s]`, inside the header/action column that sits adjacent to the `relative … aspect-[4/3]` image block. When the link mounts dynamically the flex row appears to overflow its column into/under the image; after a reload the same markup lays out correctly, which points at a dynamic-mount/layout-reflow interaction rather than plain static CSS.

Needs a browser repro to pin down (likely candidates: missing `flex-wrap` on the action row, `overflow-hidden`/stacking context of the image block, or the fadeIn animation measuring before layout). Repro: publish-capable account → generate + save a recipe → toggle Public ON → observe the View link vs the image; refresh and compare.

Related: #3262 (same toggle flow), #3208/#3209 (post-#3207 follow-ups).

adamtasteslikegood:
Closes #3262, closes #3263, closes #3264. Jira: KAN-149 (child of Sprint 3 epic KAN-136).

One branch for the three bugs from Adam's post-v0.4.4 zucchini-poppers field test — they share the SPA publish/view path and two of them share a single root cause.

## #3262 — View link on a client-predicted slug (404 / wrong page)

**Root cause (two layers):**
1. `togglePublic` (generator + recipe-detail) predicted the slug client-side (`slugFromTitle`) — but the server owns minting (`_resolve_public_slug`, collision → `-N`, KAN-139 semantics), so the prediction is wrong exactly when it matters.
2. The persistence mirror-back that should have healed it (shipped since v0.4.2, `e0346f3`) **mutated the recipe object in place** — and the app is zoneless (`provideZonelessChangeDetection`). The `currentRecipe` signal's reference never changed, so the corrected slug (and the failure revert!) never re-rendered.

**Fix:** no client-side slug derivation at all; the optimistic flip goes through the signal immutably; after the sync resolves the component adopts the server-authoritative row from auth state (the mirror-back now writes an immutable copy). The View link simply appears once the real slug exists. The KAN-104 empty-slug preflight and KAN-137 confirm-guard are untouched.

## #3263 — blank page on refresh of `/recipe/:id`

`fetchRecipeFromApi` treated the `GET /api/recipes/:id` **row shape** (`{…columns, data: {…blob}}`) as the Recipe blob → `@for` over `undefined` ingredients → blank render. Every refresh hits this path because auth isn't hydrated when the constructor's `paramMap` handler fires.

**Fix:** the KAN-139 column-over-blob merge is extracted from `PersistenceService.loadFromApi` into a shared `src/utils/recipe-row.ts` (`recipeFromRow`) — the #3209 shared-extraction argument — and the deep-link fetch now normalizes through it. Tolerates the pre-columns contract and a bare blob defensively.

## #3264 — View link paints behind the image after toggling publish ON

The action rows are **non-wrapping** flex rows inside a `minmax(0, 1fr)` grid column; a flex row can't shrink below its content, so the freshly-mounted link overflowed the half-width column and painted behind the adjacent image block (`position: relative` = own stacking context, painted on top). The generator card only shows the link right after a toggle, which is why a refresh always "healed" it. Fix: `flex-wrap` (+ `min-w-0`) on the action row and publish cluster in both templates.

## Verification

- `npm run lint` / `format:check` / `type-check` / `build` — clean
- `npm test` — 189/190; the one failure is `server/redirects.test.ts` favicon 404, **pre-existing** (fails identically on a clean `origin/dev` tree in this environment; CI was green on v0.4.4) and untouched by this diff
- New Vitest coverage: `recipe-row` unit suite (7 cases: column-over-blob, null-column clears, sourceSlug/origin fallbacks, legacy contracts); recipe-detail regressions for row normalization on cold deep link (#3263), no-client-slug + server-slug adoption (#3262), signal-driven failure revert
- #3264 is a static-CSS fix verified by analysis; worth a quick eyeball in the walkthrough: publish-capable account → generate → toggle Public ON → link should wrap below the buttons instead of clipping under the image

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,33 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ---
 
+## [Unreleased]
+
+### Fixed
+
+- **SPA publish-state fix cluster** (KAN-149,
+  [#3262](https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/issues/3262),
+  [#3263](https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/issues/3263),
+  [#3264](https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/issues/3264)),
+  from Adam's post-v0.4.4 zucchini-poppers field test:
+  - The View link no longer renders a client-predicted slug that could 404
+    or point at another recipe on a name collision — the client stops
+    deriving slugs on publish entirely and adopts the server-minted slug
+    (collision → `-N` suffix) once the sync resolves. Root cause was
+    zoneless change detection: the publish flow mutated the recipe object
+    in place, so neither the server's corrected slug nor the failure revert
+    ever re-rendered; the flip now goes through the signal immutably.
+  - Refreshing `/recipe/:id` no longer renders a blank page: the cold
+    deep-link fetch treated the `GET /api/recipes/:id` row shape
+    (`{…columns, data: {…blob}}`) as the recipe blob, leaving
+    ingredients/instructions undefined. The KAN-139 column-over-blob merge
+    is extracted into a shared `recipeFromRow` util (per #3209) used by both
+    the persistence sync and the deep-link fetch.
+  - Toggling publish ON no longer pushes the freshly-mounted View link
+    behind the recipe image: the action rows were non-wrapping flex rows
+    inside a `minmax(0, 1fr)` grid column, so the extra link overflowed
+    under the adjacent `position: relative` image block; the rows now wrap.
+
 ## [0.4.4] - 2026-07-24
 
 ### Fixed
--- a/src/components/generator/generator.component.html
+++ b/src/components/generator/generator.component.html
@@ -72,8 +72,11 @@
             Vegan Recipe
           </div>
 
-          <!-- Action Buttons -->
-          <div class="flex items-center gap-2">
+          <!-- Action Buttons — must wrap (#3264): a non-wrapping flex row
+               can't shrink below its content, so when the View link mounts
+               on publish it overflowed the half-width grid column and
+               painted behind the adjacent (position: relative) image. -->
+          <div class="flex flex-wrap items-center gap-2 min-w-0">
             <button
               (click)="onSaveRecipe()"
               [disabled]="isSaved()"
@@ -160,7 +163,7 @@
             </button>
 
             <!-- v0.2 Anti-Recipe Distribution Controls -->
-            <div class="flex items-center gap-2 ml-2 pl-4 border-l border-stone-200">
+            <div class="flex flex-wrap items-center gap-2 ml-2 pl-4 border-l border-stone-200">
               @if (canPublish()) {
                 <div class="flex items-center gap-2">
                   <span
--- a/src/components/generator/generator.component.ts
+++ b/src/components/generator/generator.component.ts
@@ -241,21 +241,34 @@ export class GeneratorComponent {
       return;
     }
 
-    recipe.is_public = nextState;
-
-    if (nextState && !recipe.slug) {
-      recipe.slug = this.slugFromTitle(recipe.name);
+    // KAN-149 (#3262): the flip is immutable and goes through the signal —
+    // zoneless change detection never re-renders on in-place mutation, which
+    // both hid the failure-revert below and froze the View link on a
+    // client-predicted slug. The slug itself is server-minted (collision →
+    // -N suffix), so no client-side guess: the View link appears once the
+    // sync merges the server's answer back.
+    const updated: Recipe = { ...recipe, is_public: nextState };
+    if (this.recipe()?.id === recipe.id) {
+      this.recipe.set(updated);
     }
 
     try {
-      const synced = await this.persistenceService.saveRecipe(recipe);
+      const synced = await this.persistenceService.saveRecipe(updated);
       if (!synced) {
         throw new Error('Publish state failed to sync to the server');
       }
+      // Adopt the server-authoritative row (slug included) into the viewed
+      // signal — auth state was just updated by the save's mirror-back.
+      const fresh = this.authService.currentUser()?.savedRecipes.find((r) => r.id === recipe.id);
+      if (fresh && this.recipe()?.id === recipe.id) {
+        this.recipe.set(fresh);
+      }
     } catch (err) {
       console.error('Failed to toggle public state:', err);
-      recipe.is_public = !nextState;
       this.authService.saveRecipe(recipe);
+      if (this.recipe()?.id === recipe.id) {
+        this.recipe.set(recipe);
+      }
       // KAN-104 (#3146): the revert already worked; without a message the
       // user just sees the switch snap back with no explanation.
       this.toastService.show(
--- a/src/components/recipe-detail/recipe-detail.component.html
+++ b/src/components/recipe-detail/recipe-detail.component.html
@@ -38,8 +38,11 @@
             Vegan Recipe
           </div>
 
-          <!-- Action Buttons -->
-          <div class="flex items-center gap-2">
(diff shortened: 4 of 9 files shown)


adamtasteslikegood:
## Review — PR #3265 (KAN-149)

### Overview
Bug-fix cluster closing #3262/#3263/#3264 (Adam's post-v0.4.4 field test). Three independent fixes on one branch, sharing the SPA publish/view path:

- **#3262 (wrong/404 View link):** `togglePublic` in both `GeneratorComponent` and `RecipeDetailComponent` no longer predicts a slug client-side. The publish flip now goes through `this.recipe.set({...recipe, is_public: nextState})` (immutable) instead of mutating the passed `Recipe` in place — the real root cause, since the app is zoneless and an in-place mutation never re-renders the `currentRecipe` signal. After sync, the component adopts the server-authoritative row (slug included) from `authService.currentUser().savedRecipes`.
- **#3263 (blank page on refresh):** the `GET /api/recipes/:id` row envelope (`{...columns, data: {...blob}}`) was rendered raw. The existing column-over-blob merge from `PersistenceService.loadFromApi` is extracted into `src/utils/recipe-row.ts` (`recipeFromRow`) and reused by the cold deep-link fetch.
- **#3264 (View link under image):** `flex-wrap` + `min-w-0` added to the non-wrapping action rows in both templates.

### Correctness
Traced the async ordering between `PersistenceService.saveRecipe()` and the component's post-sync "adopt fresh row" read:
- `AuthService.saveRecipe()` is a plain synchronous signal write (no async work), so by the time `persistenceService.saveRecipe(updated)` resolves, `authService.currentUser().savedRecipes` already reflects any server-minted slug mirrored back by `_apiSaveRecipe`. The `.find(r => r.id === recipe.id)` lookup right after is not racing anything — good.
- The failure-revert path is actually cleaner than before: since the original `recipe` param is never mutated, `catch` can just do `authService.saveRecipe(recipe)` / `this.recipe.set(recipe)` to restore true pre-toggle state, vs. the old manual `recipe.is_public = !nextState` flip-back.
- `recipeFromRow`'s `isRecipeRow` heuristic (`'data' in payload && !!payload.data`) is safe — confirmed `Recipe` (`recipe.types.ts`) has no `data` field, so a bare blob can never be misidentified as a row envelope.
- The `recipeFromRow` extraction out of `loadFromApi` is a faithful lift, not a rewrite — same column-over-blob semantics (including the "columns win, even when null" and pre-columns-contract fallback), just shared with the new call site.

### Minor / non-blocking
1. `togglePublic` is still ~45 lines duplicated verbatim between `generator.component.ts` and `recipe-detail.component.ts`, and this PR grows that duplicate rather than shrinking it. Understandable to keep this PR surgical, but worth a follow-up ticket to extract into a shared injectable, since the two copies will keep drifting otherwise.
2. The post-sync "adopt fresh row" step silently no-ops if the recipe isn't found in `savedRecipes` (edge case) — the View link would then stay on the optimistic value with no slug until next reload. Looks unreachable in practice since `auth.saveRecipe` always upserts by id, just flagging the implicit invariant.
3. #3264 is explicitly "verified by analysis" only per the PR description — agree with the author's own note that it's worth an actual eyeball (flex-wrap/stacking-context fixes are easy to get subtly wrong outside the exact repro width) before merge.

### Security / risk
No new endpoints or input-handling changes — purely reworks how the SPA renders existing Flask-published contracts. No concerns.

### CI
All `pr-gate` checks green (build, lint, type-check, tests, CodeQL, SEO, dependency review). `npm test` 189/190 with the one pre-existing failure correctly called out as unrelated. Worth noting: Codex hit its usage limit and Junie failed to run on this PR, so this may be the only completed automated review pass.

**Verdict:** logic checks out, no blocking issues. Recommend the manual click-through for #3264 the author already flagged, then good to merge.

---

---

PR #3270, opened 2026-07-25
docs(architecture): correct Valkey's role — Flask caches into it too [KAN-150]

adamtasteslikegood:
## Why

Follow-up to #3269. Adam caught that the doc I merged there described Valkey as Express's rate-limit store and nothing else — he recalled Memorystore actually caching content for Backend after the cert issue and the memory spikes were resolved. He's right; verified in code.

## What was missing

`Backend/app.py:132-197` wires Flask-Caching with `CACHE_TYPE = "RedisCache"` against Memorystore over GCP IAM auth (`utils/valkey_auth.create_iam_redis_client`), degrading to per-worker `SimpleCache` when unreachable — the state the service ran in before the Memorystore CA fix landed. One instance, two independent consumers, non-overlapping key spaces.

## Now documented

- **Both roles**, with the Flask edge added to the SPA+API diagram and a dedicated section.
- **What is cached today** — recipe image bytes, 24h TTL, key `vgc:img:<recipe_id>` (`generation_api_bp.py:290-321`). Read-through in front of GCS, legacy base64-in-DB as fallback, serving both the SPA and public SSR pages (varying only `Cache-Control`).
- **Why the cache read sits below the ownership check** — the key is global, so serving on a hit before checking access would leak private and deleted recipes' images to anyone holding the UUID.
- **Silent degradation** — falling back to `SimpleCache` is by design and invisible from behaviour; it's a log question (`Response cache: Valkey/Redis backend` vs. `in-memory SimpleCache fallback`).
- **What is *not* cached yet** — `cache_utils.py` defines keys and TTLs for recipe JSON (600s), stats (300s), and collections, and `worker_api_bp.py` already invalidates recipe entries on write, but `recipes_api_bp.py` and `collections_api_bp.py` contain no cache references at all. The invalidation side is wired ahead of the read side. Flagged as an observation, not changed here.
- **A trap warning** — `invalidate_cache()` in the legacy `recipes_bp.py` / `generation_bp.py` comes from `repositories.recipe_repository` and clears a file-based cache on the old JSON storage path. Unrelated to Valkey.
- Deployment topology now notes both services and the migrate Job reach Cloud SQL and Memorystore over private IP in the same VPC.

## Verification

`npx prettier --check` passes. Every claim traced to source: `Backend/app.py`, `Backend/utils/cache_utils.py`, `Backend/utils/valkey_auth.py`, `Backend/blueprints/generation_api_bp.py`, `Backend/config.py`, plus a negative check that `recipes_api_bp.py` and `collections_api_bp.py` reference the cache zero times.

Docs-only. No code, dependency, or submodule pointer change.

https://claude.ai/code/session_013Nx6PfdmTFXgg51amQM69f

---

### DIFF
--- a/docs/architecture/diagrams/frontend-backend.md
+++ b/docs/architecture/diagrams/frontend-backend.md
@@ -36,9 +36,13 @@ graph LR
     P -->|rows| F
     F -->|JSON| E
     E -->|JSON| B
-    V[(Valkey)] -.->|rate-limit state| E
+    V[(Valkey — Memorystore)] -.->|rate-limit state| E
+    V -.->|response cache| F
     B -.->|localStorage written first| B
 ```
+
+Valkey serves **two independent consumers**: Express's rate limiter and Flask's
+response cache. See [Valkey is shared, and Flask caches into it](#valkey-is-shared-and-flask-caches-into-it).
 
 The initial document is a **static shell**. `server/index.ts:166-168` answers the
 SPA catch-all with `dist/index.html` off disk — no template engine, no data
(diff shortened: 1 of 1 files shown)


adamtasteslikegood:
The "not cached yet" observation in this PR is now tracked as **KAN-151**.

Filing it turned up more than the PR body says: those four endpoints were not merely never wired — commit `48d5c14` ("feat: add Valkey-backed response caching with Flask-Caching", 2026-03-21) explicitly implemented caching for `GET /api/recipes/<id>`, `/api/recipes/stats`, `/api/collections`, and `/api/collections/<id>` alongside the image endpoint, and that commit is an ancestor of current `dev`. Only the image cache survives today.

No commit removes those lines — `git log --full-history -S` on the path shows only the addition — so it looks like collateral from a merge resolution in that window rather than a deliberate revert. I did not establish attribution and stopped rather than guess; KAN-151 says so explicitly.

One consequence worth calling out: `invalidate_recipe` still runs on every recipe write (`worker_api_bp.py:526,695-696`) and deletes `recipe_key` / `recipe_stats_key` entries that nothing populates, so it is currently a no-op.

No behaviour change here — this PR remains docs-only, and the doc text describing the present state stays accurate either way.

Issue #3254, opened 2026-07-24
KAN-140 - Independent second opinion via Codex (gpt-5.5, high reasoning effort) on the diff. Gate: PASS (no P1s). 2 advisory [P2] findings below — both spot-checked against the diff before posting.


adamtasteslikegood:
Independent second opinion via Codex (gpt-5.5, high reasoning effort) on the PR #3252 (KAN-140) diff. **Gate: PASS (no P1s).** 2 advisory [P2] findings, both spot-checked against the diff before posting — tracked as child issues:

- #3255 — [P2] publish-toggle `title` tooltip sits on the `disabled` button, so the "Manually entered recipes can't be published." explanation may never surface (`generator.component.html:177`, `recipe-detail.component.html:79`)
- #3256 — [P2] manual-entry recipes still write free text into the legacy `notes` field, which the KAN-140 editor renders read-only forever with no migration into `personalNotes` (`manual-entry-modal.component.ts:127`, `generator.component.ts:155`)

_Posted by Claude on Adam's behalf via `/codex review`_
--codex

_Originally posted by @adamtasteslikegood in https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/pull/3252#pullrequestreview-4776833264_

PR #3271, opened 2026-07-25
fix(spa): make an unavailable publish toggle explain itself, and free manual notes [KAN-143] [KAN-144]

### LINKED ISSUE #3255: Codex reiview KAN-140 - P2 - 1
**[P2] Advisory** — Codex (gpt-5.5, high reasoning) inline finding on PR #3252 (KAN-140).

**`src/components/generator/generator.component.html:177`** (same pattern: `src/components/recipe-detail/recipe-detail.component.html:79`)

```html
<button
  type="button"
  role="switch"
  [attr.aria-checked]="r.is_public"
  aria-labelledby="public-toggle-label"
  (click)="togglePublic(r)"
  [disabled]="
    publishToggleKind(r) === 'locked' ||
    publishToggleKind(r) === 'manual' ||
    publishTogglePending()
  "
  [title]="publishToggleTitle(r)"
  class="relative inline-flex h-5 w-9 items-center rounded-full transition-colors focus:outline-none shadow-inner"
  [class.bg-green-600]="r.is_public"
  [class.bg-stone-300]="!r.is_public"
  [class.opacity-40]="publishToggleKind(r) !== 'normal' || publishTogglePending()"
  [class.cursor-not-allowed]="
    publishToggleKind(r) === 'locked' || publishToggleKind(r) === 'manual'
  "
>
```

[P2] The explanatory `title` is attached to the same `<button>` that becomes `disabled` for manual recipes. Disabled buttons often don't surface native hover/focus tooltips, so the promised "Manually entered recipes can't be published." explanation can be invisible. Same pattern in `recipe-detail.component.html:79`. Suggest moving the `title`/tooltip affordance onto a wrapper element, or adding an explicit visible/accessible description outside the disabled control.

--codex

_Originally posted as an inline review comment on PR #3252: https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/pull/3252#discussion_r3648228151 — parent review #3254, sibling finding #3256._

### LINKED ISSUE #3256: KAN-140 Codex Review - P2 - 2
**[P2] Advisory** — Codex (gpt-5.5, high reasoning) inline finding on PR #3252 (KAN-140).

**`src/components/recipe-detail/recipe-detail.component.ts:145`** (editor seeds from `personalNotes` only; same pattern `src/components/generator/generator.component.ts:155`)

```ts
startEditNotes() {
  const r = this.recipe();
  if (!r) return;
  // KAN-140: the editor only ever touches personalNotes — the generated
  // notes render on the public page and must not be user-writable.
  this.editedNotes.set(r.personalNotes || '');
  this.isEditingNotes.set(true);
}
```

**`src/modals/manual-entry/manual-entry-modal.component.ts:127`** (manual entry still writes free text into the old `notes` field, unchanged by the PR)

```ts
const newRecipe: Recipe = {
  id: crypto.randomUUID(),
  name: info.name,
  description: info.description,
  prepTime: info.prepTime,
  cookTime: info.cookTime,
  servings: info.servings,
  ingredients,
  instructions: this.manualInstructions(),
  notes: info.notes,
  ...
```

[P2] The notes editor now always seeds from and saves to `personalNotes` only, but manual-entry recipes still write free text into the old `notes` field at creation (`manual-entry-modal.component.ts:127`, unchanged by this PR). Once a manual recipe is opened, its original notes render read-only forever, and the pencil opens an empty `personalNotes` with no way to migrate/merge the old text. Same pattern in `generator.component.ts:155`. Suggest seeding `personalNotes` from `notes` for manual-origin recipes (or have manual-entry write directly into `personalNotes` going forward).

--codex

_Originally posted as an inline review comment on PR #3252: https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/pull/3252#discussion_r3648228154 — parent review #3254, sibling finding #3255._

adamtasteslikegood:
Sprint 3 wave-2 item **B2** — the two Codex P2 findings from the KAN-140 review (parent #3254).

Closes #3255
Closes #3256

## #3255 — the explanation was attached to a control that can't show it

The publish toggle carried its reason-why in a `title` on a button that was `[disabled]` for exactly the cases the title explains. A disabled button takes no focus and shows no reliable native tooltip, so *"Manually entered recipes can't be published."* could never actually reach the user.

- `[disabled]` → `[attr.aria-disabled]` in both templates: the state is still announced, but the control stays hoverable and focusable so the `title` works.
- `togglePublic()` refuses the change and **toasts the reason** instead of returning silently. The manual case already toasted; canonical (`locked`) and mid-sync (`pending`) did not, and now use the same copy as the tooltip.

The guards in `RecipeViewBase.togglePublic()` were already the real backstop (they predate this and are tested) — the template's `disabled` was belt-and-braces on top of them, so dropping it changes no publish behaviour.

## #3256 — manual entry wrote the chef's notes into the read-only field

`manual-entry-modal.component.ts` wrote free text into `notes`, the field KAN-140 froze as generated, publicly-rendered, read-only content. A manual recipe's own notes were unreachable forever and the pencil opened an empty box.

- Manual entry now writes **`personalNotes`** (label updated to "Chef's Notes (private)").
- An **existing** manual recipe adopts its legacy `notes` into `personalNotes` on first edit and clears the legacy field, so the same text doesn't render twice. Manual recipes can never be published, so nothing public depends on their `notes`.
- Generated recipes are untouched — pinned by a test.
- Follow-on found in the browser: after migration the read-only block fell through to *"No notes added yet."* printed directly above the user's own notes. The fallback now only shows when there are no personal notes either.

## Tests

TDD — 5 new cases in `recipe-view.base.test.ts`, all watched failing for the missing behaviour first, plus a **new test file for `ManualEntryModalComponent`**, which had none.

`npm test` → 19 files, **212 passed**. Lint, `type-check`, `format:check`, and `build` all clean.

## Browser verification

Templates can't be unit-tested here (no TestBed harness), so this was verified against `npm run dev` with a seeded manual recipe carrying legacy notes:

| Check | Result |
|---|---|
| Manual toggle attributes | `aria-disabled="true"`, `disabled: false`, `title="Manually entered recipes can't be published."` |
| Activating it | toast renders: *"Manually entered recipes can't be published."* |
| Toggle focusable | `document.activeElement` reaches it (a `disabled` button cannot) |
| Notes editor on a legacy manual recipe | seeds with `skillet must be screaming hot` (was empty) |
| After save | `{"notes":"","personalNotes":"skillet must be screaming hot — and use bacon fat"}`, renders under "My notes (private)" |
| Generated recipe (control) | toggle `aria-disabled="false"`, title "Publish this recipe", generated notes still render |

One thing worth knowing: Playwright's actionability check treats `aria-disabled="true"` as disabled and refuses to click, so the activation above was dispatched with `el.click()` (which no-ops on a natively `disabled` button — so it's still a fair A/B). Real users and AT are unaffected; future browser tests targeting these toggles will need `force`.

---

### DIFF
--- a/src/components/generator/generator.component.html
+++ b/src/components/generator/generator.component.html
@@ -573,9 +573,14 @@
               </div>
             </div>
           } @else {
-            <p class="text-yellow-900 italic whitespace-pre-wrap">
-              {{ r.notes || 'No notes added yet.' }}
-            </p>
+            <!-- GH #3256 (KAN-144): "No notes added yet." directly above the
+                 user's own notes reads as a contradiction — a migrated manual
+                 recipe has no generated notes but plenty of personal ones. -->
+            @if (r.notes || !r.personalNotes) {
+              <p class="text-yellow-900 italic whitespace-pre-wrap">
+                {{ r.notes || 'No notes added yet.' }}
+              </p>
+            }
             @if (r.personalNotes) {
               <p
                 class="mt-3 pt-3 border-t border-yellow-100 text-yellow-900/80 italic whitespace-pre-wrap text-sm"
--- a/src/components/generator/generator.component.html
+++ b/src/components/generator/generator.component.html
@@ -172,15 +172,20 @@
                     >Public</span
                   >
                   <!-- KAN-139: greyed while the initial server sync is pending
-                       and for copies saved from a public recipe; disabled
-                       outright when the recipe is canonical (server-locked). -->
+                       and for copies saved from a public recipe; unavailable
+                       when the recipe is canonical (server-locked).
+                       GH #3255 (KAN-143): aria-disabled, not disabled — a
+                       disabled button takes no focus and shows no reliable
+                       tooltip, so the `title` explaining *why* could never
+                       reach the user. togglePublic() refuses the change and
+                       toasts the reason instead. -->
                   <button
                     type="button"
                     role="switch"
                     [attr.aria-checked]="r.is_public"
                     aria-labelledby="public-toggle-label"
                     (click)="togglePublic(r)"
-                    [disabled]="
+                    [attr.aria-disabled]="
                       publishToggleKind(r) === 'locked' ||
                       publishToggleKind(r) === 'manual' ||
                       publishTogglePending()
--- a/src/components/recipe-detail/recipe-detail.component.html
+++ b/src/components/recipe-detail/recipe-detail.component.html
@@ -74,15 +74,20 @@
                     >Public</span
                   >
                   <!-- KAN-139: greyed while the initial server sync is pending
-                       and for copies saved from a public recipe; disabled
-                       outright when the recipe is canonical (server-locked). -->
+                       and for copies saved from a public recipe; unavailable
+                       when the recipe is canonical (server-locked).
+                       GH #3255 (KAN-143): aria-disabled, not disabled — a
+                       disabled button takes no focus and shows no reliable
+                       tooltip, so the `title` explaining *why* could never
+                       reach the user. togglePublic() refuses the change and
+                       toasts the reason instead. -->
                   <button
                     type="button"
                     role="switch"
                     [attr.aria-checked]="r.is_public"
                     aria-labelledby="public-toggle-label"
                     (click)="togglePublic(r)"
-                    [disabled]="
+                    [attr.aria-disabled]="
                       publishToggleKind(r) === 'locked' ||
                       publishToggleKind(r) === 'manual' ||
                       publishTogglePending()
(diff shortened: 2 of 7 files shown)


adamtasteslikegood:
## Codex review (independent second opinion)

**GATE: PASS** — no blocking issues found.

Codex ran the actual test suite, lint, and type-check against this diff (7 files, generator/recipe-detail templates, `recipe-view.base.ts`, manual-entry modal + tests) rather than just reading it, plus traced the `notes` → `personalNotes` migration logic across all touched files. All green. The only failure it hit was `ng build` needing outbound network access to fetch Google Fonts — an artifact of Codex's network-sandboxed review environment, not a code defect.

**Recommendation:** Ship as-is — Codex independently verified the notes-migration and publish-toggle-explains-itself changes with zero findings, and the sole failure (font fetch) doesn't reflect a real issue with the code.

---
_Posted by Claude on Adam's behalf, via `/codex review`_

Issue #227, opened 2026-04-15
shutdownValkey(): the Promise.race timeout uses setTimeout without clearing/unref’ing it when quit() succeeds. That leaves a pending timer for up to 3s


adamtasteslikegood:
shutdownValkey(): the Promise.race timeout uses setTimeout without clearing/unref’ing it when quit() succeeds. That leaves a pending timer for up to 3s (can delay tests/process shutdown and keeps the event loop alive unnecessarily). Consider storing the timeout handle and clearing it in a finally block (and/or calling unref() on the timer) when quit resolves first.

https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/pull/174#discussion_r3074821512

PR #3272, opened 2026-07-25
feat(release): thin release-train slice + hardened git helpers [KAN-138]

adamtasteslikegood:
The mechanical half of the two-repo release train, plus a review pass over the git helpers it builds on.

## 1. The thin slice

| Piece | Runs | Does |
|---|---|---|
| `scripts/release/train-verify.sh` | anywhere | five read-only drift stations; exit **0** clean / **1** blocking / **2** could-not-determine; `--json`, `--for-release` |
| `.github/workflows/release-train.yml` | Actions — dispatch + daily | runs the verifier, publishes a job summary, fails on blocking drift |
| `scripts/release/train-backsync.sh` | **local only** | opens/merges the `main → dev` back-sync PRs; merge commit enforced, dry run by default |

Stations: cookbook + Backend `main→dev` back-sync debt, `dev→main` promotion debt, submodule pointer (SHA **and** tree — a promotion merge changes the SHA but not the content), alembic head count, version/CHANGELOG coherence.

**It caught real drift on its first run**: Backend `main` carries `dfc5cad` (the v0.4.4 promotion merge, #245) never back-synced to `dev`. Third occurrence in the record after #227/#236. Trees identical, so no code is missing — it is ancestry debt, which is exactly what the station exists to surface.

The write half stays local on purpose: `GITHUB_TOKEN` is scoped to this repo and cannot open the Backend PR, and **`dev` carries `required_linear_history` in both repos**, so a merge-commit merge into `dev` only lands for an actor with ruleset bypass. Squashing instead would defeat the point — it rewrites the commits, ancestry never converges, and the drift count never returns to zero.

## 2. The `action_required` gate — the pilot's diagnosis was wrong

KAN-138 recorded that dev→main release runs sit at `action_required` and the automation must clear them via the runs-approve API. It does not.

Every run ever held in that state, on either repo, is the **Junie** workflow triggered by **Copilot** on `pull_request_review*` events. On `dev`: zero runs ever held. On `main`: four, all Junie/Copilot. The cause is a repo policy — `fork-pr-contributor-approval` is set to `all_external_contributors`, stricter than GitHub's default — which holds runs whose triggering actor is not a collaborator. The release path's own gate jobs run as the repo owner and are never held, and Junie is not a required check. The 7 approvals during the pilot were unnecessary.

Verified ruleset facts are recorded in `scripts/release/README.md` and on KAN-138.

## 3. Hardened git helpers

Adversarial review by parallel subagents: **47 findings raised, 10 refuted, 37 confirmed by reproduction and fixed.** The load-bearing ones:

- **`git-workflow.sh` could publish a dangling submodule pointer.** The submodule push was gated on *this run* having created a commit, so an already-committed-but-unpushed Backend was skipped while `git add -A` still staged the moved gitlink. Reproduced end to end on scratch repos: parent pushed, pointer referencing a SHA that exists only locally, exit 0, "All operations completed successfully!". Cloud Build would fail to check that SHA out at release time. Push now fires whenever origin lacks the submodule HEAD, and `assert_submodules_published()` hard-aborts before the parent stages anything.
- **`ahead-behind.sh` silently reported the parent repo's branches under the Backend heading** when the submodule was uninitialized — the exact trap CLAUDE.md warns about. Now detected via the gitlink and failed loudly.
- **`ahead-behind.sh` returned `0 0` for an unresolvable `--base`** — a false clean bill of health — and never consumed its own per-repo status code.
- `--dry-run` mutated the index; `--auto` could commit the error string as the message; a failed `git commit` was unchecked; colour output went to stdout and polluted commit messages; `-m` silently re-enabled push after `--commit-only`; prompts hung in non-TTY runs.

**Scope correction on top of the agents' work:** their protected-branch pre-flight refused on branch identity alone, which would have blocked the most common submodule-spanning workflow here — the pointer bump, where `Backend/` sits clean on `dev` (or detached, where `git submodule update` leaves it) and only the parent has work. It now skips submodules with nothing to commit and nothing to push, so the guard fires on what is about to be written. Verified both directions.

`commit-phase-1.sh` deleted — a one-shot 2024 script hardcoding `git add requirements.txt` (removed in Backend #215). **This is the one non-surgical change in the PR**; easy to revert if you'd rather keep it.

## Verification

- `bash -n` clean on both scripts; `--help` exits 0 on both
- `ahead-behind.sh --base dev . Backend` (the CLAUDE.md invocation) — sane output, exit 0
- uninitialized-submodule fixture → `❌ Uninitialized submodule`, exit 1
- `--dry-run --all` leaves the index untouched (verified before/after)
- protected-branch guard: clean Backend proceeds, dirty Backend on `dev` refused, `--allow-protected` overrides
- `train-verify.sh` and `train-backsync.sh` both smoke-tested against the live repos

**Known gap, not fixed here:** a rejected `git commit` still leaves the overall exit status 0 (nothing is pushed). Needs a run-level failure flag.

The workflow can't be test-dispatched until this is on `dev` — GitHub only lists `workflow_dispatch` from the default branch.

---


adamtasteslikegood (commented):
## Overview

This PR does three things: (1) a read-only "release train verify" station (`train-verify.sh` + a scheduled/dispatch Actions workflow) that checks back-sync debt, promotion debt, submodule pointer state, alembic head count, and CHANGELOG coherence; (2) a local-only `train-backsync.sh` that opens/merges the `main→dev` back-sync PRs; (3) a substantial hardening pass over `scripts/git/git-workflow.sh` and `scripts/git/ahead-behind.sh`, plus removal of the dead `commit-phase-1.sh`.

The hardening pass is the bulk of the diff and it's good work — real, reproducible bugs (dangling submodule pointer, `ahead-behind.sh` silently reporting the parent's branches under an uninitialized submodule, `--dry-run` mutating the index, unchecked `git commit` exit status, prompts hanging on non-TTY, `-m` silently re-enabling push, etc.) with fixes that match the stated invariants. Below are the things worth a second look before/after merge.

## Findings

**1. `train-verify.sh`: an unresolvable pointer tree is reported as "differs" rather than "could not determine"**

```sh
pointer_tree=$(git -C Backend rev-parse "$pointer^{tree}" 2>/dev/null || echo "unknown")
...
pointer_content_state="matches-backend-main-content"
[ "$pointer_tree" = "$backend_main_tree" ] || pointer_content_state="differs-from-backend-main-content"
```

If the pinned submodule SHA's commit object isn't reachable in the local `Backend/` object database (e.g. it lived only on a since-rewritten/rebased branch and was pruned), `pointer_tree` becomes the literal string `"unknown"`, which then compares unequal to `backend_main_tree` and gets reported as `differs-from-backend-main-content` — a real blocking failure in `--for-release` mode. That's exactly the anti-pattern the file's own header calls out ("a checker that reports 'fine' when it could not actually check is how a broken gate ships green") — just inverted: here an *indeterminate* result gets reported as a *confident* block. Given `count_ahead()` a few lines above already treats an unresolvable ref as fatal (`die`), the same treatment (or at minimum a distinct `"indeterminate"` state that maps to a warning, not a hard block) would be more consistent. Low likelihood in the CI path (fetch-depth 0, both branches fetched) but plausible in ad hoc local runs against a shallow or stale `Backend/` checkout.

**2. `git-workflow.sh`: the disclosed "commit failure still exits 0" gap is real and traced through**

I verified this against the full file rather than just the diff. `commit_changes()` correctly returns 1 on a failed `git commit` (e.g., a rejected pre-commit/pre-push hook), and both call sites (`process_submodules`, `process_main_repo`) correctly gate `committed`/`MAIN_COMMITTED` on that return value, so a failed commit won't get force-pushed or falsely marked as pushed. However, neither `process_submodules` nor `process_main_repo` propagates that failure to `main()` — both are called as bare statements under `set -e`, and since the failure is only ever observed inside an `if commit_changes; then ...; fi`, the enclosing function's own exit status is whatever its last unconditional command was (a `cd`), not the failure. So `main()` reaches "✅ All operations completed successfully!" and exits 0 even though a repo's changes were never committed. This is honestly disclosed in the PR description as a known gap, but since this PR already introduces the "distinguish 1 (checked, bad) from 2 (couldn't check)" exit-code discipline for `train-verify.sh`, it'd be worth applying the same discipline here — a single `WORKFLOW_FAILED=1` flag set by `commit_changes`'s failure branches and checked at the end of `main()` (with `exit 1` instead of the success banner) would close it in a handful of lines. Not necessarily a blocker for this PR, but I'd want it tracked as a fast-follow rather than left open indefinitely, since a script that silently reports success on a failed commit is easy to trust incorrectly during a release.

## Minor / non-blocking

- `train-verify.sh`'s alembic-heads station parses `revision`/`down_revision` out of migration files with regex rather than calling `flask db heads` — a reasonable dependency-free approximation (documented as such), but it will under- or over-count if a merge migration's `down_revision` tuple ever spans multiple lines. Worth a comment noting the limitation, or falling back to the real Alembic call when a `Backend/.venv` (or similar) is available.
- `release-train.yml` runs `train-verify.sh` twice (once piped to text, once with `--json`) to get both a human summary and machine-readable output. Harmless for a daily/dispatch job, but could be one invocation with `tee` + a small `jq`/python reformat if it ever needs to run more often.
- The one intentionally non-surgical change (deleting `commit-phase-1.sh`) is well-flagged in the description and looks like reasonable dead-code removal — the script hardcoded a 2024 branch/path history — but noting it here since the PR itself invites a second opinion on it.

## Security / correctness of the new workflow

`release-train.yml` looks fine from a security standpoint: `permissions: contents: read` only, triggered only by `workflow_dispatch` and `schedule` (never `pull_request`), so it never executes against untrusted fork content. No secrets are used.

## Test coverage

No automated tests for the shell changes, consistent with this repo's existing convention for `scripts/git/*.sh` (no bash test harness in place). The PR's manual verification list (`bash -n`, `--help`, dry-run index check, uninitialized-submodule fixture, live smoke tests against both scripts) is a reasonable substitute given that.

---

adamtasteslikegood:
**Known gap now closed** (`826fdb7`) — and it was worse than the PR body said.

The rejected-commit case was only half of it. None of the four `git push` invocations in `push_changes` checked their status either, so a fatal push printed `✅ Pushed <repo> to origin/<branch>` directly over git's own `fatal: Could not read from remote repository`, and the run still exited 0.

Both are fixed by a run-level failure flag. The part that needed care is that a non-zero return from `commit_changes` already meant two different things:

| rc | meaning | sets the flag? |
|---|---|---|
| 1 | benign no-op — nothing to commit, user declined at a prompt | no |
| 2 | requested and did not happen — commit rejected, staging failed, push failed or refused | **yes** |

Without that split, every quiet run would fail: the pointer-bump case (submodule clean, nothing to commit) returns 1 constantly and is entirely normal. The run now ends by naming each operation that did not complete and exits 1.

Push *refusals* — protected branch, detached HEAD, branch/target mismatch — count as failures too. The push was asked for and did not happen; a caller should not read that as success.

Verified on scratch repos:

| case | exit |
|---|---|
| pre-commit hook rejects the commit | **1**, naming `commit in the main repository` |
| unreachable remote (push fails) | **1** |
| happy path | 0 |
| nothing to commit | 0 |

And against this repo: dry-run pointer bump → 0, dirty `Backend/` on `dev` → 1, index untouched either way.

adamtasteslikegood:
Worked through all five points. Two fixed in `5fb2f92`, one was already closed in `826fdb7`, one I'm pushing back on, one is Adam's call.

**1. Unresolvable pointer tree reported as "differs" — fixed.**

Correct, and it's the file's own stated sin inverted: an indeterminate result dressed up as a confident block. Worse than described, actually — in `--for-release` the message it printed was *"promote Backend dev→main first"*, which would not fix anything, since the real cause is a missing object locally. Now dies with the actual remedy (`git -C Backend fetch origin`), matching how `count_ahead()` already treats an unresolvable ref.

**2. `git-workflow.sh` exiting 0 after a failed commit — already fixed in `826fdb7`**, pushed before this review landed. Your trace of the mechanism is right: the failure is only ever observed inside `if commit_changes; then …; fi`, so the enclosing function's status is its last unconditional command (`cd`), and `main()` sails on to the success banner.

The implemented fix is the `WORKFLOW_FAILED` flag you describe, plus one thing that turned out to be necessary: a non-zero return from `commit_changes` already meant two different things, and conflating them would have made every quiet run fail. Split into `1` = benign no-op (nothing to commit, user declined a prompt) and `2` = requested and did not happen. Only `2` sets the flag. The pointer-bump case returns `1` constantly and is entirely normal.

It also covered a hole this review didn't reach: none of the four `git push` calls checked their status either, so a fatal push printed `✅ Pushed <repo> to origin/<branch>` on top of git's own `fatal: Could not read from remote repository` — and still exited 0.

**3. Alembic regex vs a multi-line `down_revision` tuple — fixed, and it's not theoretical.**

Built a four-migration fixture with a wrapped merge tuple. Old pattern: **3 heads reported where there is 1** — a hard block on a release that is fine. The `(.+)$` capture stopped at the newline, so the merge's parents never got recorded and both merged heads stayed "live". Pattern now matches a parenthesised or bracketed group across lines. Kept the parse rather than shelling out to `flask db heads` so the station needs no Backend virtualenv; noted the reasoning in a comment.

**4. Running the verifier twice in the workflow — leaving as is.**

The second call is `--no-fetch` against local refs — sub-second, read-only, no network. Merging them would mean deriving the human table from the JSON, which is reimplementing the renderer in `jq`, or `tee`-ing and parsing, which couples the exit code to output formatting. Keeping them independent means a change to the text rendering can't corrupt the signal the job fails on. Worth revisiting if this ever runs per-push rather than daily.

**5. `commit-phase-1.sh` deletion** — flagged for exactly that second opinion. It hardcodes `git add requirements.txt`, a file removed in Backend #215, plus `PHASE_1_*.md` paths, so running it today stages nothing meaningful. Adam's call; a revert is one command.

On test coverage: agreed there's no bash harness here, but every fix in this PR was verified by reproduction on scratch fixtures rather than by inspection — including both bugs above, where the fixture is what showed the failure was real.

PR #3275, opened 2026-07-25
feat(release): pointer targets Backend main, spent-version + CHANGELOG-SHA gates, freeze spec [KAN-138]

adamtasteslikegood:
Adds the last station Adam identified for `--for-release`, and specs the freeze
window for the driver that does not exist yet.

## The station: spent version

`release.yml` checks `git ls-remote --tags` first and gates every subsequent
step on the result:

```
release.yml:58   if git ls-remote --exit-code --tags origin "refs/tags/$TAG"; then exists=true
release.yml:70   - name: Extract release notes      if: exists == 'false'
release.yml:87   - name: Create and push tag        if: exists == 'false'
release.yml:100  - name: Publish GitHub Release     if: exists == 'false'
```

So merging a `dev → main` release PR while `package.json` still names an
already-tagged version yields a green PR, a green workflow, **no tag, no
Cloud Build trigger, no deploy**, and nothing anywhere reporting a failure.
`package.json` is on `0.4.4` and `v0.4.4` is tagged — the trap is armed today.

`--for-release` now blocks on it. It queries origin even under `--no-fetch`
(a single ref lookup, the same query `release.yml` makes, so the two cannot
disagree) and `die`s rather than guessing when the query itself fails: a wrong
"not tagged" here greenlights a release that silently never ships.

```
BLOCK v0.4.4 is already tagged — release.yml would skip the tag, the Release,
      and the Cloud Build trigger and report success. Bump package.json
      (patch, at minimum) before cutting.
```

## The freeze spec

Docs only — `scripts/release/README.md`. Two findings worth recording:

1. **A `dev → main` PR merges the live tip of `dev`**, not the tip as of when
   the PR was opened. So the freeze has to open when the version bump lands on
   `dev`, not when the release PR opens — by then it is already too late.

2. **`update` ("Restrict updates") does not do the job.** It governs direct
   pushes, and both branches already block those by requiring a PR, so it adds
   nothing against the actual threat: a *merge*. The primitive that blocks
   merges is a required status check that never reports.

The unlock is asymmetric: `dev` frees at tag-cut (that is what makes "new push
= patch bump" true, and the back-sync needs `dev` writable), `main` stays held
until the deploy verifies — because when Cloud Build fails, the only correct
recovery is a patch bump through the normal train, which is exactly what v0.3.0
and v0.3.4 did as v0.3.1 and v0.3.5.

Enforcement is specced but **unverified** — it needs the ruleset created and one
throwaway PR to confirm the block lands. Detection (SHA-pinning the four refs at
freeze-open) is the layer to build first: no GitHub config written, nothing to
clean up after a crashed run.

## Verification

- `bash -n` clean; both output modes exercised; `--json` parses.
- Default mode still exits 0 on the current tree; `--for-release` exits 1 with
  the new BLOCK line.

---

### DIFF
--- a/scripts/release/README.md
+++ b/scripts/release/README.md
@@ -93,6 +93,65 @@ Two options, neither blocking:
 - **Relax to `first_time_contributors`** if the pending-run noise is annoying:
   `gh api -X PUT repos/{o}/{r}/actions/permissions/fork-pr-contributor-approval -f approval_policy=first_time_contributors`
 
+## The freeze window (spec — for when the driver lands)
+
+Today's scripts verify and back-sync; they do not drive the train. When a driver
+does exist it needs a freeze, because the train's payload is mutable underneath
+it: **a `dev → main` PR merges the live tip of `dev`, not the tip as of when the
+PR was opened.** Anything merged to `dev` between opening the release PR and
+merging it ships silently, outside the CHANGELOG, under a version that never
+described it.
+
+### When it opens
+
+The moment the version bump + CHANGELOG land on `dev`. That is when `dev`'s tip
+*becomes* the release payload — earlier than the release PR, which is already
+too late.
+
+### When it closes — asymmetric, and this is the part worth getting right
+
+| ref | unlocks at | why there |
+| --- | --- | --- |
+| `dev` (both repos) | **tag `vX.Y.Z` exists on origin** | The tag is the moment the version becomes immutable, which is precisely what makes "any new push means a patch bump" true. It also *has* to be here: step 8/9 back-sync writes to `dev`. |
+| `main` (both repos) | **deploy verified live** | If Cloud Build fails, the only correct recovery is a patch bump through the normal train. Holding `main` through that window is what stops a panic hotfix pushed straight to it — the one move that would break `main === deployed`. |
+
+Not a hypothetical: v0.3.0 died in the migrate job and v0.3.4 died parsing the
+Express Dockerfile. Both recovered exactly this way, as v0.3.1 and v0.3.5.
+
+### Mechanism — two layers
+
+**1. Detection (build this first).** At freeze-open the driver snapshots the four
+ref SHAs — `dev` and `main` on both repos — and re-checks them before every
+mutating step. Any movement aborts and names the ref that moved. No GitHub
+config is written, so a crashed run leaves nothing to clean up, and it catches
+the real failure at the moment it would do damage.
+
+**2. Enforcement (needs a decision + one experiment).** Note what does *not*
+work: the `update` rule ("Restrict updates") governs direct pushes, and both
+branches already block those by requiring a PR — so it adds nothing against the
+actual threat, which is a **merge**. The primitive that blocks merges is a
+**required status check that never reports**: a dedicated `release-freeze`
+ruleset per repo, normally `enforcement: disabled`, holding one
+`required_status_checks` rule for a context (`release-freeze`) that no workflow
+ever publishes. Flip it to `active` to freeze; every PR merge then sits at
+"expected — waiting for status", except for bypass actors. Bypass stays repo
+admin in `pull_request` mode, so the train's own merges still land while other
+agent sessions, Dependabot merges, and UI merges are held.
+
+This layer is unverified here — it needs the ruleset created and one throwaway
+PR to confirm the block actually lands. It also needs a standalone `--unfreeze`
+escape hatch, and the freeze state written locally *before* the API call, or an
+aborted train leaves both repos frozen with no record of why.
+
+### Already enforced
+
+The "spent version" half of the rule is live now, as a station:
+`--for-release` refuses to proceed when `v<package.json version>` is already
+tagged. `release.yml` checks `git ls-remote --tags` first and skips tag
+creation, the GitHub Release, and therefore the Cloud Build trigger — so
+re-merging a release PR on a spent version produces a green PR, a green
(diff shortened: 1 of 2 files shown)


adamtasteslikegood (commented):
## Overview

Two changes: a new "spent version" station in `scripts/release/train-verify.sh --for-release`, and a docs-only spec for the freeze window in `scripts/release/README.md`. The spent-version check closes a real gap — `release.yml` gates tag/Release/Cloud-Build-trigger creation on `git ls-remote --tags` finding no existing tag, so re-merging a release PR on an already-tagged `package.json` version currently produces a green PR and workflow with silently no deploy.

## Code quality / correctness

- **`set -euo pipefail` compatibility is handled correctly.** `git ls-remote --exit-code ... || ls_remote_rc=$?` and the `git rev-parse --verify --quiet ...` inside the `if` are both in conditional contexts, so a non-zero exit doesn't trip `set -e`. `$?` is captured from the right command in the `||` idiom.
- **Exit-code trichotomy is right and matches the file's own contract** (0/1/2 exit codes documented at the top): `0` (tag found) → `already-tagged`, `2` (`--exit-code`'s "no match" code) → `not-tagged`, anything else → `die` rather than assume "safe to ship". This is exactly the failure mode the PR is trying to close, applied consistently to itself.
- **`tag_state` is assigned on every path** (case's three arms + both branches of the local-mode `if/else`), so no unset-variable risk under `set -u`, and the verdict-section comparison (`[ "$tag_state" != "already-tagged" ]`) only ever runs where `tag_state` came from the `--for-release` branch, so it can't be fooled by the `(local)`-suffixed value from the other mode.
- **`--no-fetch` semantics get quietly narrower**: the new station always queries `origin` over the network even when `--no-fetch` is passed, which is called out explicitly in both the usage text and the code comment, so it's a documented tradeoff rather than a hidden surprise. Worth flagging anyway since it's the one place the flag no longer means "no network."
- Consistent with the file's existing idioms (die() for exit 2, comment-explains-the-why style, printf-based JSON building).

## Risk / blast radius

`train-verify.sh` is invoked by `.github/workflows/release-train.yml` (dispatch + daily cron), which fails the job on blocking drift but isn't a required PR-gate check — so a bug here would surface as a bad daily/dispatch run, not a merge block. Low blast radius. Note the workflow calls the script twice per run when `for_release=true` (once for text, once for `--json`), so the new `git ls-remote` network call doubles — pre-existing pattern, not introduced by this PR, but worth keeping in mind if it ever gets flaky.

## Suggestions

- No automated test exercises the new tag-check branch (both `--for-release` and local-tag paths) — consistent with the rest of the file having no test harness, so not blocking, just noting the verification described in the PR body (`bash -n`, manual exercise of both modes) is manual rather than repeatable.
- Minor: the doc's "Already enforced" section could link back to the specific station number/comment in `train-verify.sh` for anyone reading the README first.

## Security

No injectable input — `$version` comes from `package.json` via `node -p`, is only ever passed as a quoted argument (never eval'd), and the ref name is used as-is by `git ls-remote`/`git rev-parse`, so no shell-injection surface here.

Nothing blocking. Nice, self-consistent close of a real silent-failure gap.

---

adamtasteslikegood:
## Added: pointer targets Backend `main` (`45d0138`)

Per Adam's call — the submodule pointer pins **Backend `main`'s own SHA**, not a
`dev`-side SHA that merely carries main's content.

Both deploy identical code, so this is not about what runs. It is about
answering "which Backend commit is in production?" from one ref:

| release | pointer pinned | Backend `main` was | trees |
| --- | --- | --- | --- |
| v0.3.9 | `18a303a` | `53af0941` | identical |
| today | `50cdbbb` | `dfc5cad` | identical |

Each time, the deployed SHA appears nowhere on the branch that is meant to
equal production.

`--for-release` now blocks unless `pointer == Backend origin/main`, and splits
the message by cause because the fix differs:

- **identical trees** → pointer aimed at the wrong SHA. Bump it (the message
  prints the exact command).
- **differing trees** → `dev → main` never promoted. Promote, then pin.

The `dev` comparison is retained as a diagnostic in the table and JSON.

## Added: CHANGELOG must name the pinned SHA

Production deploys whatever the pointer pins at tag time, so without this the
release notes describe the frontend half and stay silent about the other half.
The convention already existed informally (v0.3.7, v0.3.9, v0.4.0) — this makes
it checkable and accepts any abbreviation, matching how those entries read.

Section matching is scoped heading-to-next-h2, not file-wide: a file-wide grep
matches the *previous* release's pointer line and reports present. Verified:

```
0.3.9 + 18a303a (its own pointer)  -> present
0.4.4 + 18a303a (prior section)    -> absent          <- no cross-section bleed
0.4.4 + dfc5cad                    -> absent
9.9.9 + anything                   -> missing-section
```

## Current state on `dev`

`--for-release` exits 1 with all three gates firing, each naming its own fix:

```
BLOCK submodule pointer pins 50cdbbb891ae, Backend main is dfc5cad816b6
      (identical tree) — bump the pointer to main's own SHA:
      git -C Backend checkout dfc5cad816b6 && git add Backend
BLOCK CHANGELOG section for 0.4.4 does not name the pinned Backend SHA (50cdbbb891ae)
BLOCK v0.4.4 is already tagged — release.yml would skip the tag, the Release,
      and the Cloud Build trigger and report success.
```

Default mode still exits 0 (all three are warnings there), so the daily
scheduled run does not go red on a repo that is simply between releases.

_Posted by Claude on Adam's behalf_

PR #3276, opened 2026-07-25
docs: single-source CLAUDE.md, stub AGENTS.md, unstale the Valkey issue refs [KAN-153]

adamtasteslikegood:
`CLAUDE.md:294` and `AGENTS.md:262` have described GH #163/#162 as **open** for
three months. Both closed `COMPLETED` on **2026-04-15**, fixed by `042e8e5`
*"fix(valkey): catch shutdownValkey() failures to prevent reinitialization from
aborting"* — literally #163's title — which is on `origin/main` and in released
tags.

## Why this is worth a PR

The stale line caused a live misdiagnosis today. This morning's automated
health-check routine reported:

> `redis.exceptions.AuthenticationError` on flask-backend (106 occurrences over
> the 24h window, first seen 2026-07-20). This matches the already-tracked
> Valkey/rate-limiter auth issue referenced in the repo docs (GH #163/#162,
> KAN-16/KAN-17) … it's a known issue already tracked in the backlog.

It is not that issue. #163/#162 were **Express-side** (`server/valkey.ts`)
connection-teardown edge cases. The live error is **Flask-side** IAM
token-refresh auth, which is what Backend #247 addresses. The doc line turned a
new production error into a closed-backlog non-finding.

So the replacement text names what *is* live and where, rather than just
deleting the dead reference:

- Backend #247 — IAM token-refresh auth failures
- KAN-151 — the response-cache code lost in merge `07123c2`

Docs only; no code paths touched.

_Opened by Claude on Adam's behalf_

---
Rovo Dev code review: <strong>Rovo Dev has reviewed this pull request</strong>
Any suggestions or improvements have been posted as pull request comments.

Issue #163, opened 2026-04-10
Catch quit failures before reinitializing Valkey client


adamtasteslikegood:
> **<sub><sub>![P1 Badge](https://img.shields.io/badge/P1-orange?style=flat)</sub></sub>  Catch quit failures before reinitializing Valkey client**
> 
> When the singleton guard detects an unhealthy existing client, it calls `shutdownValkey()` directly inside the `catch` path; if `client.quit()` rejects (common when the socket is already broken), that rejection escapes `createValkeyClient()` and prevents the intended reinitialization/fallback path from running. This means a transient Redis failure during a second initialization attempt can now bubble up as a hard failure instead of recovering.
> 
> Useful? React with 👍 / 👎. 

 _Originally posted by @chatgpt-codex-connector[bot] in [#109](https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/pull/109/changes#r2996193418)_

PR #3277, opened 2026-07-25
docs: agent-instruction hygiene — single-source CLAUDE.md, fix broken instructions across all five files [KAN-153]

adamtasteslikegood:
Continues KAN-153. #3276 merged carrying only its first commit (the stale
`#163/#162` fix); these three were pushed to that branch afterwards and never
entered it. Same branch content, new PR.

## 1. AGENTS.md → stub (297 → 58 lines)

It claimed to be "kept in sync for the core sections" with `CLAUDE.md`. It was
drifted in four places — each verified against the code, not against
`CLAUDE.md`'s wording:

| AGENTS.md said | reality |
| --- | --- |
| `blueprints/generation_bp.py` serves `/api/generate` | it's `generation_api_bp.py:42,117`; `generation_bp.py` only serves `/generate_recipe` (HTML form helper) |
| "Dual-auth: Flask tries user OAuth first" | both live call sites pass `None` — `services/image_service.py:84`, `blueprints/worker_api_bp.py:604` — so generation runs on the server key. Only `POST /api/models/refresh` forwards session creds (`blueprints/api_bp.py:62`) |
| model names "include `models/` prefix" | prefix applies to the **list API**; `config.py DEFAULT_MODEL` and the generation paths use bare IDs. Both forms live |
| "Prettier runs as a CI job and commits fixes on push" | `CLAUDE.md`: `ci.yml` is a push-only near-no-op safety net. **Flatly contradictory** |

Cause: yesterday's KAN-130 pass corrected `CLAUDE.md` five times (`f7b3d8d`,
`29bfedc`, `bf1bec3`, `3e53ef4`, `12b44d1`) and never touched `AGENTS.md`, whose
last edit was the first of those.

**Nothing dropped.** Six sections were AGENTS-only but universal rather than
harness-specific, so they moved *into* `CLAUDE.md`: commit-and-push cadence, the
`npm run pm:*` loop, `sync_jira_confluence_status.py` deps/env, and the Jira
project-key allowlist (KAN/RCP writable, PLZG/TO read-only, TOSVC frozen).
Verified present after the move.

## 2. GBrain blocks removed from CLAUDE.md (491 → 394 lines)

Lines 394–491 were machine state, not repo instructions: gbrain `0.42.56.0`,
`~/.gbrain/config.json`, the Railway Postgres engine, embedding model, artifacts
repo, "134 pages, 831 chunks", the 0.18.x rollback backup. One laptop's config
in a public repo, every number rotting on the next reconfigure — the same class
KAN-130 was trimming when `12b44d1` pulled a point-in-time page count from this
very section.

The Search Guidance half is worse than stale: it sits between
`<!-- gstack-gbrain-search-guidance:start/end -->` and is regenerated by
`/sync-gbrain` from a template, so hand-written text inside it is destroyed on
the next run — including the paragraph warning about that, which the file
acknowledged and then placed inside the block anyway.

**Kept** as a Non-obvious patterns bullet: `Backend/` is a separate
non-federated gbrain source, queries need `--source gstack-code-backend` or they
silently miss, and a bare `/sync-gbrain` from inside `Backend/` registers a
duplicate source. Repo hazard verified with `--dry-run` on 2026-07-24 (KAN-130),
not machine config.

## Caveats

- `/sync-gbrain` will re-add its guidance block on the next run unless the skill
  is told not to write here. This deletion is not durable on its own.
- `prettier` cannot run locally (Node 26.5 vs the pinned build throws
  `SyntaxError` in `legacy-cli.mjs`, pre-existing), so
  `Frontend — lint + format` on this PR is the real formatting verification.

Backend half: adamtasteslikegood/tasteslikegood.com#248

---


adamtasteslikegood (commented):
## Review

**What this PR does:** Collapses `AGENTS.md` from 297→58 lines into a stub that defers to `CLAUDE.md`, migrates six universal-but-AGENTS-only sections (commit/push cadence, `npm run pm:*` loop, status-script deps, Jira project-key allowlist) into `CLAUDE.md`, and drops the two trailing GBrain blocks (~97 lines of machine-specific state) from `CLAUDE.md`. Docs-only, 2 files changed.

**Correctness — spot-checked every factual claim in the PR body against the actual code (Backend submodule initialized to verify):**
- ✅ `/api/generate` is served by `generation_api_bp.py:42` (`@generation_api_bp.route("/generate", ...)`); `generation_bp.py` only serves `/generate_recipe` (line 261, HTML form helper). Confirmed.
- ✅ `get_genai_client(session_credentials)` in `services/gemini_service.py:22` — dual-auth fallback (session creds → `GOOGLE_API_KEY`) matches the new CLAUDE.md wording exactly.
- ✅ `config.py:55` — `DEFAULT_MODEL = "gemini-3.1-pro-preview"` — bare ID, no `models/` prefix. Confirmed the "both forms in active use" nuance is accurate.
- ✅ `ci.yml` — `push`-triggered only, auto-commits Prettier fixes on diff. Matches "push-only auto-commit safety net," and does flatly contradict the old AGENTS.md framing ("Prettier runs as a CI job... don't be alarmed by bot commits" reads as routine, which it isn't given branch protection blocks direct pushes to `dev`/`main`).
- ✅ The gbrain/Backend-submodule bullet was correctly retained in "Non-obvious patterns" rather than dropped with the rest of the GBrain content — it's repo-hazard documentation, not machine state, and the PR body's distinction holds up.

No factual regressions found — every claim I could verify against source matched.

**Process notes (informational, not blocking):**
- The Junie review bot posted "Junie is failed! Junie output file path is not set" — looks like a tooling/infra issue on that workflow, unrelated to this diff. `Gate — all checks passed`, `CodeQL`, and `prettier` are all green.
- The Atlassian acceptance-criteria bot flagged the GH #163/#162 unstale-reference item as not met — but that update (`091d2b3`) already landed in `dev` via #3276 before this branch's three new commits; it's not in this PR's diff because it didn't need to be. Worth a quick note back to the bot/issue so it doesn't stay flagged red.
- Minor: PR title/branch carry `KAN-153`, but the three commit messages in this PR say `[KAN-16]`. Given Jira scans the PR title (per this repo's own convention doc), the title is what matters and it's correct — but the commit-message mismatch is worth a glance in case it's tracking the wrong issue.

**Risk:** Low. Markdown-only change, no code paths touched, no test coverage implications. The stated caveat that `/sync-gbrain` will re-add the Search Guidance block on next run is honestly disclosed rather than silently left as a landmine.

---

adamtasteslikegood:
## Scope extended: all five agent-instruction files, not just AGENTS.md

This repo tracks **five**: `CLAUDE.md`, `AGENTS.md`, `COPILOT.md`, `GEMINI.md`,
`.github/copilot-instructions.md`. Auditing the rest turned up two instructions
that fail if followed — these are not stylistic fixes.

### 1. `copilot-instructions.md:241` — directive that silently no-ops

> "If you add a new variable that genuinely needs to be read in the Angular
> client bundle, use the `VITE_` prefix — Vite will inline it at build time."

False. This is an Angular CLI build; it does not expose `import.meta.env`, and
`grep -rn "import.meta.env" src/ server/` returns **0 matches**. A variable
added that way reads `undefined` in the browser, with no build error to catch
it. Replaced with what actually works (`src/environments/`) plus an explicit
note about the silent failure.

Adam had independently deleted the milder version of the same claim from
`COPILOT.md`; that edit is folded in here, and `src/environments/environment.prod.ts`
carried the last `VITE_` reference in the repo — a comment describing an
override path that does not exist.

### 2. `GEMINI.md:43` — command that fails today

```bash
pip install -r requirements.txt
```

`Backend/requirements.txt` was deleted in Backend #215 on **2026-07-18**. The
project is on `uv sync`. Corrected, along with the "managed with `pip` and
`requirements.txt`" line above it. `setup_venv.sh` in the same block does still
exist, so only the two wrong lines changed.

### Also in this PR

`2231e83` (claude[bot]) removed AGENTS.md's pointer to the GBrain sections that
this same PR deletes from CLAUDE.md — correct catch, kept.

### Verified

- `prettier` clean on all four touched files (via the Node API; the CLI is
  broken on Node 26.5, see the earlier commit)
- `tsc --noEmit` clean
- Backend submodule pointer deliberately **not** staged — the local checkout sits
  on a feature branch and this is docs-only

### Follow-up worth having

Every defect in this PR is a doc claim that was true when written and rotted
silently. A pr-gate check could catch most of the class mechanically: resolve
every `#N` issue reference and fail on closed ones, and assert that files named
in fenced commands still exist. `gh-aw`'s `doc-updater` would not — it is
change-driven over the last 24h of merged PRs, and none of these bullets had
been touched recently.

_Posted by Claude on Adam's behalf_

PR #3283, opened 2026-07-25
chore(seo): promote orange-chicken seitan recipe to canonical [KAN-158]

adamtasteslikegood:
## What

Adds `vegan-orange-chicken-style-seitan-with-white-rice` as the **eighth** canonical recipe:

- `specs/canonical-recipes.json` — new entry, `updated` bumped to 2026-07-25
- `index.html` — matching `<noscript>` anchor
- `specs/canonical-recipes.schema.json` — `maxItems` 7 → 8
- `scripts/seo/check_canonical_recipes.sh` — count bound 3–7 → 3–8

Adam's pick. It is the set's first Asian/takeout-style entry, and it already earns its keep: a partial-slug Google query returns it as the top result with its hero image intact.

## On raising the cap

The schema and the gate both capped the list at 7, so both move together — leaving either behind makes the two disagree about what is valid. The cap exists to keep the `<noscript>` nav a *curated* set rather than a dump of the catalog, so it moves deliberately, one slug at a time, with approval on the PR. A comment in the script records that and points at the schema as the thing to keep in sync.

## Verification

Checked against the gate's own criteria **before** adding the slug: HTTP 200, self-canonical, Recipe JSON-LD with 17 ingredients / 10 steps / image / yield 4.

```
$ scripts/seo/check_canonical_recipes.sh --live https://www.tasteslikegood.org
OK: .../r/classic-vegan-margherita-pizza → 200
... (8 slugs, all 200)
OK: 8 canonical recipes validated (JSON ↔ index.html consistent)
```

Scored 5/5/5/5 — the search-intent and content-quality scores are observed rather than estimated (live ranking; ingredient/step counts from the page's JSON-LD).

## Ships with

Backend PR #251 (`is_canonical = true` for this slug), which is the half that actually makes the page durable: the repository guards then reject unpublish, re-slug, and delete, so the public page survives the SPA copy being deleted later. **Merge #251 first**; a submodule pointer bump follows.

---

### DIFF
--- a/index.html
+++ b/index.html
@@ -139,6 +139,7 @@
         <a href="/r/classic-vegan-chocolate-chip-cookies">Classic Vegan Chocolate Chip Cookies</a>
         <a href="/r/maple-smoked-tempeh-blt">Maple Smoked Tempeh BLT</a>
         <a href="/r/vegan-street-style-tofu-tacos">Vegan Street Style Tofu Tacos</a>
+        <a href="/r/vegan-orange-chicken-style-seitan-with-white-rice">Vegan Orange Chicken Style Seitan with White Rice</a>
       </nav>
     </noscript>
   </body>
--- a/scripts/seo/check_canonical_recipes.sh
+++ b/scripts/seo/check_canonical_recipes.sh
@@ -2,7 +2,7 @@
 # CI gate: validate canonical recipe list against index.html anchors.
 # Reads specs/canonical-recipes.json and checks:
 #   1. JSON is valid
-#   2. 3–7 recipes listed (Phase 0/1 bound)
+#   2. 3–8 recipes listed (Phase 0/1 bound)
 #   3. Every slug has a matching <a href="/r/{slug}"> in index.html <noscript>
 #   4. No anchors in index.html that aren't in the canonical list
 #
@@ -37,10 +37,13 @@ if ! jq empty "$CANONICAL_FILE" 2>/dev/null; then
   exit 1
 fi
 
-# 2. Count check (3–7)
+# 2. Count check (3–8). Raised from 7 for KAN-158: the upper bound exists to keep
+# the noscript nav a curated set rather than a dump of the whole catalog, so it
+# moves deliberately, one slug at a time, with Adam's approval on the PR. Keep
+# this in sync with maxItems in specs/canonical-recipes.schema.json.
 count=$(jq '.recipes | length' "$CANONICAL_FILE")
-if (( count < 3 || count > 7 )); then
-  echo "FAIL: canonical list has $count recipes (must be 3–7)"
+if (( count < 3 || count > 8 )); then
+  echo "FAIL: canonical list has $count recipes (must be 3–8)"
   errors=$((errors + 1))
 fi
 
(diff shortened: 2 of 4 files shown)


adamtasteslikegood:
## Review

The diff itself is clean and self-consistent — confirmed independently:

- `specs/canonical-recipes.json` is valid JSON with 8 entries; the new entry's `],`/`}` insertion reads oddly in the raw diff (git matched the old last-entry's closing brace as context against the new one) but the resulting file parses correctly.
- Count bound raised consistently in both places that need it: `scripts/seo/check_canonical_recipes.sh` (3–7 → 3–8) and `specs/canonical-recipes.schema.json` (`maxItems` 7 → 8).
- `index.html` `<noscript>` anchor added 1:1 with the new slug.
- CI is fully green (`Gate — all checks passed`, CodeQL, Dependency Review, prettier), and the existing automated Claude review already validated this same consistency.

## One gap none of the automated reviews caught

This PR's companion, **Backend PR [#251](https://github.com/adamtasteslikegood/tasteslikegood.com/pull/251)** (sets `is_canonical = true` for `vegan-orange-chicken-style-seitan-with-white-rice`, which per its own description is what makes the row undeletable/un-unpublishable — the actual durability guarantee for an indexed `/r/` page), is still **open, unmerged**. This cookbook PR doesn't touch the `Backend` submodule pointer at all.

Per this repo's release model (`CLAUDE.md` § Branching strategy), that's fine for merging into `dev` — no hard technical coupling in this diff, nothing here calls into the backend. But it does mean: if `dev` → `main` gets promoted and released before #251 merges into Backend `dev` *and* a follow-up PR bumps the `Backend` submodule pointer to pick it up, production would list this recipe as canonical (nav, `specs/canonical-recipes.json`) without the backend's `_guard_canonical` protection actually being in place — the exact "indexed page turns into a 404" scenario #251's description warns about.

Recommend holding the release (not the `dev` merge) until: #251 merges → `git submodule update --remote Backend` → pointer-bump PR merges. Worth a checklist note or linked-PR comment so it doesn't get lost between now and the next release cut.

---

adamtasteslikegood:
Confirmed on the sequencing — that's the right call, and it's the reason the PR description says "**Merge #251 first**". Making it an explicit checklist here so it can't get lost:

1. Merge Backend **#251** → Backend `dev`
2. `git submodule update --remote Backend`
3. Pointer-bump PR → cookbook `dev`
4. Only then cut `dev` → `main`

Two things I verified while checking the claim:

**The CI gate is not a safety net here.** `.github/workflows/pr-gate.yml:150` runs `check_canonical_recipes.sh` with no `--live` flag, so it validates JSON ↔ `index.html` consistency only and never fetches the URLs. A canonical slug that had gone dead would pass CI silently. (I ran `--live` manually against production for this PR — all 8 slugs 200.)

**One refinement to the risk framing.** Releasing this ahead of the pointer bump would not itself produce a 404: `/r/vegan-orange-chicken-style-seitan-with-white-rice` is already public and serving 200 today, independent of `is_canonical`. What's actually missing in that window is the *guarantee* — `_guard_canonical` isn't in the deployed backend, so the row stays deletable and unpublishable while the nav advertises it as canonical. The 404 is the downstream consequence if someone then deletes or unpublishes it, not an immediate effect of the release. Same conclusion, just a slower-burning failure than "released → broken".

Worth noting the same window already exists for the seven slugs seeded by `c8d2f6a1e9b3` — this PR doesn't widen it, it just adds an eighth row to the set that depends on it.

PR #3285, opened 2026-07-25
ci(release): make the single-Alembic-head check a required gate [KAN-138]

adamtasteslikegood:
## The gap

The single-Alembic-head check **already existed** and was already correct — it is station 4 of `scripts/release/train-verify.sh`. Nothing was wrong with the logic.

What was missing: **nothing ran it before a merge.**

| runs it | trigger | blocks a merge? |
|---|---|---|
| `release-train.yml` → `train-verify.sh` | `workflow_dispatch` + daily 15:00 cron | no |
| `pr-gate.yml` (the required check) | every PR | **had no Alembic check at all** |

So a PR bumping the `Backend` submodule pointer onto a branched-head Backend passed `Gate — all checks passed`, merged clean, and the drift surfaced on the next cron run — or at deploy time, which is worse: two heads make `flask db upgrade` refuse, which fails the `flask-backend-migrate` Cloud Run Job and aborts `cloudbuild.yaml` mid-release with the previous Flask revision left serving.

## Changes

- **`scripts/release/alembic-heads.sh`** (new) — the head-counting logic lifted verbatim out of `train-verify.sh`, so there is one implementation instead of two that drift.
  - bare mode → prints the count (or `error`), always exits 0. That is the contract `train-verify.sh`'s `$(...)` capture already relied on, so its BLOCK line still renders.
  - `--check` → prints a verdict, names every head with its file, exits 1 unless there is exactly one.
  - Still pure `python3` parsing. No Backend virtualenv, no `uv sync`, no `DATABASE_URL`, no app import. A gate that needs a database to tell you the database migration is malformed is a gate that gets skipped.
- **`train-verify.sh`** — calls the shared script instead of carrying its own copy.
- **`pr-gate.yml`** — new `Backend — single Alembic head` job, listed in `gate.needs` so it is part of the required `Gate — all checks passed` status check.

## Verification — including that it can fail

A gate that only ever passes is decoration, so the negative cases were run explicitly:

| case | result |
|---|---|
| real tree | `OK: Alembic has exactly one head.` → `e4a7b2d9c5f1`, exit 0 |
| synthetic 2nd head | `FAIL: Alembic has 2 heads` — names both heads + files, prints the `flask db merge` command, **exit 1** |
| missing submodule | `FAIL: could not determine Alembic heads` — points at `submodules: true`, **exit 1** |
| `train-verify.sh --no-fetch` | still reports `alembic heads 1`, exit 0, no blocking drift |

Also: `pr-gate.yml` parses as YAML, `alembic-heads` appears in both `jobs` and `gate.needs`, Prettier clean, `bash -n` clean on both scripts.

## Note on branch protection

This adds a job to the `gate` aggregate, which is already the required check — so it takes effect without touching branch-protection settings. No ruleset change needed.

https://claude.ai/code/session_01LqwHC5WeVAtgosDb469Foy

---

### DIFF
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -107,6 +107,22 @@ jobs:
           PUBSUB_EMULATOR_HOST: localhost:8085
         run: uv run pytest
 
+  # Branched Alembic heads make `flask db upgrade` refuse to run, which fails
+  # the flask-backend-migrate Cloud Run Job and aborts the deploy mid-release.
+  # This ran only on release-train.yml's manual dispatch and daily cron, so a
+  # pointer bump onto a two-head Backend could merge clean and surface hours
+  # later. It is a required check now: same script train-verify.sh station 4
+  # calls, so the gate and the release train cannot disagree.
+  alembic-heads:
+    name: Backend — single Alembic head
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: recursive
+      - name: Verify exactly one Alembic head
+        run: ./scripts/release/alembic-heads.sh --check
+
   # ── Docker ────────────────────────────────────────────────────────────────
   # Builds the production Express image (no push, no credentials). Catches
   # Dockerfile/lockfile breakage before a release tag hands it to Cloud Build
@@ -160,6 +176,7 @@ jobs:
       - frontend-build
       - frontend-test
       - backend-test
+      - alembic-heads
       - docker-build
       - changelog-check
       - canonical-recipes
--- a/scripts/release/alembic-heads.sh
+++ b/scripts/release/alembic-heads.sh
@@ -1,0 +1,123 @@
+#!/usr/bin/env bash
+#
+# Alembic head count for the pinned Backend submodule.
+#
+# Why this is its own script rather than an inline block: two callers need the
+# same answer at different moments, and a second implementation would drift.
+#
+#   1. scripts/release/train-verify.sh — station 4, at release time.
+#   2. .github/workflows/pr-gate.yml   — the required status check, on every PR.
+#
+# (2) is the one that actually blocks a merge. Before it existed, the check ran
+# only on release-train.yml's manual dispatch and its daily 15:00 cron, so a PR
+# that moved the submodule pointer onto a branched-head Backend merged clean and
+# the drift surfaced hours later — or at deploy time, which is worse: two heads
+# make `flask db upgrade` refuse, the flask-backend-migrate Cloud Run Job fails,
+# and cloudbuild.yaml aborts the release mid-flight with the old revision left
+# serving.
+#
+# Heads are parsed out of the migration files rather than shelled out to
+# `flask db heads` on purpose: that keeps this runnable with nothing but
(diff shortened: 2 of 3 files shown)


adamtasteslikegood (commented):
## Independent review

### Overview
Extracts the Alembic head-counting logic out of `train-verify.sh` into a standalone `scripts/release/alembic-heads.sh`, and wires it into `pr-gate.yml` as a new required `alembic-heads` job feeding the `gate` aggregate. Closes the real gap described in the PR body: the check previously only ran on `release-train.yml`'s manual dispatch / daily cron, so a submodule-pointer bump onto a branched-head Backend could merge clean and only surface at deploy time when `flask db upgrade` refuses to run.

### Verification performed
Rather than take the PR body's test-case table on faith, I re-ran the negative cases against a scratch `ALEMBIC_VERSIONS_DIR` and against the real `Backend/migrations/versions` tree (13 files, flat, single head today):

- Real tree → `OK: Alembic has exactly one head.` (`e4a7b2d9c5f1`), exit 0 — matches CI's green `Backend — single Alembic head` run.
- Synthetic 2-head fixture → `FAIL: Alembic has 2 heads`, names both + files, prints the `flask db merge` command, exit 1.
- Missing/nonexistent `ALEMBIC_VERSIONS_DIR` → `FAIL: could not determine Alembic heads`, points at the `submodules: true` checkout requirement, exit 1.
- Bare mode (`alembic-heads.sh` with no args) on both the 2-head and missing-dir fixtures → prints `2` / `error` respectively, **always exits 0** — matches the contract `train-verify.sh`'s `$(...) || echo "error"` capture depends on.
- A merge migration whose `down_revision = (\n    "a",\n    "b",\n)` tuple is wrapped across lines (the exact scenario called out in the code comment) → correctly parses both parents and reports 1 head, not 3.
- Typed annotations (`revision: str = ...`) parse correctly too.

All of the above match the PR description's claims exactly. `bash -n` is clean on both scripts.

### Code quality
- Single-implementation fix is the right call — the old copy in `train-verify.sh` and this one could not drift now that there's one script both callers shell out to.
- Exit-code contract (bare mode always 0, `--check` mode 1 unless exactly one head) is documented in the header comment and actually honored — verified above.
- Parsing via regex instead of `flask db heads` is a deliberate, well-justified tradeoff (no venv/DB dependency in a required PR gate), and the comment explaining *why* is worth its weight — it's the kind of decision a future reader would otherwise "simplify" back into a footgun.

### Minor, non-blocking observations
- `count_heads()` and `list_heads()` each re-implement the same revision/parent regex scan as two separate inline Python blocks (one file walk each). Not a correctness issue at 13 files, but if this ever gets refactored, folding both into one parse pass that returns both the count and the head list would remove the duplication.
- `versions.glob("*.py")` is non-recursive. Not an issue against the current flat `Backend/migrations/versions/` layout, but worth keeping in mind if migrations are ever split into subdirectories.

### Risk assessment
No blocking issues found. CI already confirms this on the real tree (`Backend — single Alembic head` ✅, `Gate — all checks passed` ✅, CodeQL ✅). The new job only runs `actions/checkout` + a pure-bash/python3 script — no new secrets, network calls, or write permissions introduced.

---

PR #3300, opened 2026-07-27
feat(security): authenticate Express→Flask with a Google ID token [KAN-170]

adamtasteslikegood:
Closes the P1 in [KAN-170](https://tasteslikegood.atlassian.net/browse/KAN-170). The Jira ticket holds the forensic timeline, the live-state snapshot, and the service hostname — deliberately not repeated here, since this repo is public.

## The problem

`flask-backend` accepted unauthenticated requests from the public internet for ~4.6 months. `POST /api/generate` completed and billed Gemini/Imagen for anonymous callers, bypassing the 20/hr AI budget, the 300/15m limiters, express-validator's body validation and 10kb cap, and Helmet's headers.

`cloudbuild.yaml` could not have revealed it. `--no-allow-unauthenticated` edits the **IAM policy**, while the `run.googleapis.com/invoker-iam-disabled=true` annotation switches the **invoker IAM check off wholesale**. The annotation wins, so the flag is inert and cannot clear it — and both artifacts a reviewer would check (this file, and the IAM policy) read as *secure*.

## This PR

Express now proves who it is, so the invoker check can be turned back on.

- **`server/flask-auth.ts`** mints a Google-signed ID token via `google-auth-library` (already a dependency). Audience is `URL.origin` of `FLASK_BACKEND_URL` — verified in the installed source that `IdTokenClient.getRequestHeaders()` caches the token and decodes its JWT `exp`, refetching only inside a 5-minute threshold. Per-request calls are cache hits, so there is **no refresh timer to leak across SIGTERM** (deliberately unlike `valkey.ts`, whose timer is right for its stateful `AUTH`).
- **`server/proxy.ts`** sends it in `X-Serverless-Authorization`.
- **`scripts/gcloud/kan170_{verify,path_a,path_b}.sh`** — dry-run by default, `--apply` to mutate.
- **`docs/security/KAN-170_PUBLIC_EGRESS_REMEDIATION.md`** — the runbook.

## Three decisions worth reviewing

**1. `X-Serverless-Authorization`, not `Authorization`.** Cloud Run forwards `Authorization` to the container unmodified, and Flask reads it in two live paths: `require_admin` (guarding `/api/admin/*`, which Express proxies) and `require_pubsub_oidc`. Overwriting it would break both. The header is set *after* the `...req.headers` spread, and any client-supplied value is deleted first so a browser cannot inject its own.

**2. The audience is `URL.origin`, not the raw env var.** A trailing slash would mint `aud="https://.../"` and 401 the entire site, and nothing else in the stack validates the shape of `FLASK_BACKEND_URL`.

**3. Fail-open on a mint failure.** `getFlaskAuthHeader()` returns `null` rather than throwing, so the request still forwards. Before the cutover that means the site is completely unaffected by a broken token path; after it, Cloud Run's 403 plus a throttled error log name the cause. Local dev mints nothing (non-`run.app` target), so `npm run dev` is unchanged.

## This release changes no production behaviour

Express starts sending a token; Flask still accepts anonymous callers. Nothing about live behaviour changes until a deliberate, separately-run cutover.

That is why the `--invoker-iam-check` commit is **reverted in this PR** and will land as its own follow-up after the cutover. It is a *service-level* setting applied the instant the Flask deploy step runs — and cloudbuild deploys Flask before Express (KAN-30). Shipping it here would 403 every proxied route for the ~60s between those two steps, or persistently if the `roles/run.invoker` grant had not been applied first.

Operator sequence: `path_a.sh prepare --apply` → release this → verify the token is arriving → `path_a.sh cutover --apply` → land the follow-up PR.

## Testing

246 tests pass; type-check, lint (`--max-warnings 0`), prettier and the server build are clean. 19 new tests cover audience derivation (trailing slash, path, custom domain, localhost, malformed), client reuse, the fail-open path, and the proxy's header discipline.

Two bugs found and fixed during review of my own change: an unhandled rejection that could terminate the container, and a disconnect guard that wrongly checked `req.destroyed` — which is true for every AI-endpoint request because `validation.ts` consumes that body stream to buffer `rawBody`. Five tests caught the latter.

Note: `server/redirects.test.ts`'s `/favicon.ico` test fails in any checkout whose path contains a dot-segment (Express `sendFile` uses `dotfiles: 'ignore'`). Pre-existing and environment-only — confirmed by stashing this change and re-running; CI checks out to a clean path.

[KAN-170]: https://tasteslikegood.atlassian.net/browse/KAN-170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---


adamtasteslikegood (commented):
## Incident-response review (KAN-170)

Reviewed as *"does this close the incident"* rather than as a code review. The remediation mechanism, the ordering discipline, and the rollback path all hold up — the sequencing rationale for reverting `--invoker-iam-check` out of this PR is correct, and `X-Serverless-Authorization` is the right call given `require_admin` and `require_pubsub_oidc` both read `Authorization`.

Also confirming, since it was the point of KAN-171: **this PR does not re-leak anything.** No project number, no real hostname — test fixtures use `flask-backend-xyz.a.run.app` placeholders, and `PROJECT_ID` is the already-public project *id*, not the number.

Three findings, one of which is a factual correction that makes the incident record **better** than the PR currently claims.

### 1. The forensic gap this PR declares permanent is already closed — and the answer is clean

See the inline comment on the runbook. Short version: there is a **second log bucket at 400-day retention**, so the full exposure window is observable. I queried every generation POST from 2026-03-01 to now: **zero unattributed calls.** The exposure was never exploited.

### 2. The detection gap that caused the 4.6-month dwell is not closed

This PR fixes the vulnerability. It does not fix *why nobody noticed for 4.6 months* — which, by the PR's own root-cause analysis, is that the annotation is invisible in both artifacts a reviewer would check.

Nothing here runs on a schedule or blocks a merge. `kan170_verify.sh` is operator-run, and there is no `.github/workflows/` or `cloudbuild.yaml` change in this PR. **This repo has learned this exact lesson before:** the single-Alembic-head check existed in `scripts/` for months while never blocking anything, until it was wired into `pr-gate.yml` and `gate.needs` (KAN-138, #3285). A check in `scripts/` is not a gate.

After cutover the annotation is sticky in the *safe* direction, so this is not urgent — but it is the difference between "fixed" and "stays fixed". Suggest a follow-up asserting `invoker-iam-disabled` is absent across all services, either in pr-gate or as a scheduled workflow.

### 3. Post-cutover failure signal is thin

`getFlaskAuthHeader()` fails open and logs at most once per 60s. Pre-cutover that is exactly right. Post-cutover, a mint failure means site-wide 403s whose only breadcrumb is a throttled `console.error` with no alert attached. A log-based alert on `[FlaskAuth]` would make the fail-open safe in both directions. Minor, and arguably follow-up work.

### Verdict

No blocking objection. Findings 2 and 3 are follow-ups, not changes to this diff. Finding 1 is worth editing before merge, because the runbook is the durable incident record and it currently understates what is known.

---

docs/security/KAN-170_PUBLIC_EGRESS_REMEDIATION.md
```diff
@@ -0,0 +1,92 @@
...
+- **Both services share the default compute SA.** Granting "Express's SA" `run.invoker` therefore grants it to anything running as that SA, including flask-backend itself. Accepted here, not hidden — splitting the identities is separate work and deliberately not bundled into a security cutover.
+
+## What this does not fix
+
+- `express-frontend` is *also* `invoker-iam-disabled=true`, guarded solely by `ingress=internal-and-cloud-load-balancing`. Acceptable for a public frontend behind a load balancer, but it should be a recorded decision rather than an accident of DRS.
+- Flask still has no rate limiting of its own (`flask-limiter` is not a dependency) and no auth on the generation endpoints. Express remains the only limiter; both paths work by ensuring Express is the only reachable caller.
+- **Billing has not been reviewed.** Request-log retention reaches only 2026-06-27, so ~3.6 months of the 4.6-month exposure is unobservable. In the visible window, 139 of 140 POSTs were Express's own egress and the single external caller was Adam's workstation — absence of evidence, not evidence of absence. A Gemini/Imagen spend review by month since March is still outstanding.
```
adamtasteslikegood: **Factual correction — this is better news than written, and the conclusion should be strengthened.**

> Request-log retention reaches only 2026-06-27, so ~3.6 months of the 4.6-month exposure is unobservable.

This is not the case. `gcloud logging buckets list` shows **two** buckets — the 30-day `_Default` and a second at **400-day retention**. Flask request logs are queryable well past the stated edge:

```
2026-03 -> 2026-03-23T09:54:15Z
2026-04 -> 2026-04-18T03:27:15Z
2026-05 -> 2026-05-06T10:49:55Z
2026-06 -> 2026-06-27T20:36:35Z
```

The 2026-06-27 figure is what you get from the most-recent-N entries under a `--limit`, not the true horizon — request volume is high enough that the cap is reached long before the retention edge. Adding an explicit `timestamp>=` filter reaches the whole window.

So the gap is closable, and I closed it. Every billable-generation POST since 2026-03-01:

```bash
gcloud logging read \
 'resource.type="cloud_run_revision" AND resource.labels.service_name="flask-backend"
  AND httpRequest.requestMethod="POST"
  AND (httpRequest.requestUrl:"/api/generate" OR httpRequest.requestUrl:"/generate_recipe")
  AND timestamp>="2026-03-01T00:00:00Z"' --limit=4000
```

Every single hit resolves to Google infrastructure (Express's own egress) or to Adam's ISP range and workstation. **Unattributed callers: zero.** Across the entire 4.6-month exposure, nobody else found it.

Two consequences:

1. **"absence of evidence, not evidence of absence" can be retired.** This is now a positive finding: exposure confirmed, exploitation excluded on evidence. That is the sentence that determines whether KAN-170 closes as a *breach* or an *exposure*, so it is worth getting right in the durable record.
2. **The outstanding Gemini/Imagen spend review is no longer needed as an abuse check.** Request logs answer it directly and more precisely than cost aggregates would. Worth keeping only if you want it for cost reasons.

One small precision note while here: "the single external caller was Adam's workstation" undercounts — several distinct `136.124.x` addresses appear (dynamic ISP assignment) plus an iPad via the Pinterest in-app browser, which is where the full unauthenticated generate → poll → image chain shows up. All attributable to Adam, so the conclusion is unchanged, but "a single caller" isn't quite what the logs say.

docs/security/KAN-170_PUBLIC_EGRESS_REMEDIATION.md
```diff
@@ -0,0 +1,92 @@
...
+- **The audience must be the bare origin** — no path, no trailing slash. `server/flask-auth.ts` derives it with `URL.origin` rather than trusting the raw env var; a trailing slash would 401 the entire site.
+- **Diagnostics are inverted from intuition.** `403` = token valid but the caller lacks `run.invoker`. `401` = token missing, malformed, or wrong audience. `404` = the request arrived as EXTERNAL and was refused by internal ingress (a Path B misconfiguration, not an auth failure).
+- **Both services share the default compute SA.** Granting "Express's SA" `run.invoker` therefore grants it to anything running as that SA, including flask-backend itself. Accepted here, not hidden — splitting the identities is separate work and deliberately not bundled into a security cutover.
+
+## What this does not fix
+
+- `express-frontend` is *also* `invoker-iam-disabled=true`, guarded solely by `ingress=internal-and-cloud-load-balancing`. Acceptable for a public frontend behind a load balancer, but it should be a recorded decision rather than an accident of DRS.
```
adamtasteslikegood: Agree with the assessment, and I verified it independently today — `express-frontend` returns **404 on both hostname forms** (opaque-hash and project-number), so ingress is genuinely holding.

Worth being precise about the framing though, because it is the reusable lesson from this incident: **ingress is the differentiator, not IAM.** Both services are equally unauthenticated at the IAM layer; only reachability differs. That is exactly why `flask-backend` was exposed and this one was not.

Since this paragraph says it "should be a recorded decision rather than an accident of DRS" — that recording does not exist yet as far as I can tell. Suggest filing it as a follow-up and linking it here, otherwise the twin defect lives on only in a runbook paragraph, which is roughly how the original annotation stayed invisible for 4.6 months.

(If a ticket already exists, ignore me — but a link here would help the next reader.)

scripts/gcloud/kan170_verify.sh
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env bash
```
adamtasteslikegood, line 1: **This is the right oracle, but nothing runs it.**

The anon-403 / site-200 assertion is exactly the correct verification, and having it scripted rather than described is a real improvement. My concern is lifecycle, not content: this executes only when an operator types it, and there is no `.github/workflows/` or `cloudbuild.yaml` change anywhere in this PR.

The incident's damage was dwell time, not the misconfiguration itself — ~4.6 months during which `cloudbuild.yaml` and the IAM policy both read as secure. That is a *detection* failure, and detection is the one thing this PR does not add.

Concretely, after the cutover nothing would catch:
- the annotation being re-disabled via console or a stray `gcloud run services update`
- `express-frontend`'s ingress being widened later (its only remaining guard)
- a future service created with the same DRS-driven default

The precedent worth copying is KAN-138: the single-Alembic-head check sat in `scripts/` for months, ran only on dispatch and cron, and blocked exactly nothing until it was added to `pr-gate.yml` *and* `gate.needs` — the second half being the part that actually made it a gate.

A scheduled workflow asserting `invoker-iam-disabled` is absent across all Cloud Run services would close this, and most of the logic already exists in this file. Follow-up, not a blocker for this PR — the sequencing argument for keeping this release behaviour-neutral is sound and I would not want this bundled in.

server/flask-auth.ts
```diff
@@ -0,0 +1,118 @@
...
+  if (!audience) return null;
+
+  try {
+    if (!cached || cached.audience !== audience) {
+      const auth = new GoogleAuth();
+      cached = { audience, client: auth.getIdTokenClient(audience) };
+    }
+    const client = await cached.client;
+    const headers = await client.getRequestHeaders();
+    return headers.get('authorization');
```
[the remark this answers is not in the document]
adamtasteslikegood: The casing premise does not hold. `getRequestHeaders()` returns a WHATWG `Headers` object, and `Headers.get()` is **case-insensitive by specification** — it normalises the name before lookup. Verified empirically on this exact runtime:

```
$ node -e "const h=new Headers(); h.set('Authorization','Bearer x'); console.log(h.get('authorization'))"
Bearer x
```

So a `Authorization` vs `authorization` mismatch cannot produce the silent `null` this describes. That was the load-bearing half of the concern.

The residual half — that a `null` here is indistinguishable from the no-audience case in the logs — is real in principle, but unreachable in practice: `IdTokenClient.getRequestHeaders()` either sets `authorization` or throws, and a throw is already caught and logged by the `catch` below. Adding `if (!value) throw` would guard a branch that cannot execute, at the cost of converting a deliberate fail-open into a throw inside the one function whose contract is *never throws*. Declining on that basis.

Where the underlying instinct **is** right: post-cutover, observability of a failed mint is genuinely thin — the existing failure log is throttled to once per 60s and has no alert attached. That is worth fixing, but as a log-based alert on `[FlaskAuth]` rather than as a guard here. Raised separately in the incident-response review on this PR.

server/proxy.ts
```diff
@@ -31,11 +32,46 @@ export function createFlaskProxy(label = 'Flask') {
...
+      console.error(
+        `[${label} Proxy] ${sanitizeForLog(req.method)} ${sanitizeForLog(req.originalUrl)} dispatch failed: ${sanitizeForLog((err as Error).message)}`
+      );
+      if (!res.headersSent) {
+        res.status(502).json({ error: 'Backend service unavailable' });
+      }
+    });
+  };
+
+  async function forward(req: Request, res: Response): Promise<void> {
```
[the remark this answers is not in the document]
adamtasteslikegood: This one is incorrect for this codebase — `forward` **must** be re-created per call, because it closes over `label`, which is a parameter of `createFlaskProxy`:

```ts
export function createFlaskProxy(label = 'Flask') {   // proxy.ts:29
  ...
  `[${label} Proxy] ... failed: ...`                  // proxy.ts:78
```

That closure is the entire purpose of the factory. Hoisting `forward` to module scope would mean threading `label` through as an explicit argument on every invocation and giving up the reason the factory exists — strictly more code and more coupling to avoid allocating a function object.

On the cost itself: there are three call sites, all at module load (`/api` and SSR in `index.ts`, plus `'Auth'` in this file). That is three function objects, once, at startup — not per request. The comment's own Details section concedes this is "a one-time cost and not a hot-path issue", which I read as agreeing there is no performance argument here.

No change. If anything the current shape is the consistent one: it is a factory, and every factory in this module closes over its arguments.

docs/security/KAN-170_PUBLIC_EGRESS_REMEDIATION.md
adamtasteslikegood: Addressed in `31b6147`.

The bullet is replaced by an **Exploitation assessment** section carrying the positive finding, the reproducible query, and the explicit conclusion that KAN-170 closes as an exposure rather than a breach — plus a note that no notification obligation arises, since the anonymous surface exposes no personal data (`/api/recipes` returns 0 records unauthenticated; it is session-scoped).

I also wrote down the trap itself, because it is the reusable part: **do not infer retention from a `--limit`ed query.** Two buckets exist (`_Default` at 30d, a second at 400d), and `gcloud logging read` returns the most-recent N, so a busy service hits the cap long before the retention edge and the oldest row masquerades as the horizon. Pin the window with `timestamp>=`; `--freshness` is also unreliable alongside `--order=asc`.

One refinement while editing: the window is anchored to `2026-03-09T10:20:31Z`, the ingress-open moment your `kan170_path_b.sh` comment already identifies as the exposure event. The query starts 2026-03-01, so it covers the full period with a week of margin either side.

The spend review is now described as optional and cost-motivated rather than outstanding as an abuse check.

Separately, I added a third bullet under _What this does not fix_ recording the detection gap — nothing asserts the annotation stays absent — since that is the half of the incident this PR deliberately does not close.

adamtasteslikegood:
## Independent code review — summary

Reviewed the diff independently (token minting in `server/flask-auth.ts`, header discipline in `server/proxy.ts`, the two staged `gcloud` scripts, the verify script, and the runbook). Design and header discipline are sound; found one real risk in the proxy's new `await` that isn't covered by the review already on this thread.

### Finding: unhandled `'error'` on `req` during the token-mint `await` (streamed proxy path)

`server/proxy.ts` `forward()`:

```ts
async function forward(req: Request, res: Response): Promise<void> {
  const originalHost = req.headers.host || req.hostname;
  const rawBody = (req as RawBodyRequest).rawBody;
  const authHeader = await getFlaskAuthHeader();   // <-- yields here
  if (res.destroyed) return;
  ...
  } else {
    req.pipe(proxyReq, { end: true });             // <-- req's first listener
  }
}
```

Pre-PR, `req.pipe(proxyReq, ...)` ran synchronously in the same tick as the handler, so `req` was never listener-less (`.pipe()` attaches an internal `'error'` handler on the source). Now, for every non-AI proxied request (`/api/recipes`, `/api/collections`, `/r/*`, `/browse`, `/sitemap.xml`, `/static/*` — the majority of traffic), `req` has **zero** listeners for the full duration of `await getFlaskAuthHeader()`.

In steady state that's a microtask-only cache hit and harmless. But per the `flask-auth.ts` docstring, the underlying `IdTokenClient` does a real multi-second metadata round-trip on: the first request after every cold start, the ~5-minute eager-refresh window before each ~60-minute token expiry (once per container instance per hour), and any request right after a mint failure resets `cached`. If a client disconnects during one of those windows (tab closed, mobile drop, aborted `fetch()`), Node emits `'error'` on the request stream with no listener — which is an uncaught exception under Node's default `EventEmitter` behavior. `server/index.ts` registers no `uncaughtException` handler (only `SIGTERM`/`SIGINT`), so this would crash the whole Express process for every concurrent user, not just the disconnecting one. Notably, `forward(req, res).catch(...)` already guards against unhandled *rejections* from this same `await` — this is the analogous unhandled stream *error* that isn't covered by that guard.

Suggested fix, one line before the `await`:
```ts
req.on('error', () => {}); // keep req listener-covered across the await below
```

### What's solid (agreeing with the review already on this PR)

- Header discipline: `delete headers['x-serverless-authorization']` unconditionally, then reassign only if a real token exists — correctly blocks client injection whether auth is on or off, covered by `proxy-auth.test.ts`.
- Audience derivation via `URL.origin` + `.run.app` allowlist — correctly strips trailing slash/path/query and skips the metadata probe entirely for local dev.
- `X-Serverless-Authorization` (not `Authorization`) avoids colliding with Flask's `require_admin`/`require_pubsub_oidc`.
- Fail-open on mint failure is explicitly and correctly scoped to the pre-cutover state, not a hidden default.
- Test suite is unusually precise about *why* each case exists (audience edge cases, cache reuse, injected-header stripping, throttled failure logging).

CI is green across the board (Gate, CodeQL, pytest, Docker build, Dependency Review), and this diff doesn't touch `cloudbuild.yaml` or the invoker-IAM annotation, consistent with the PR's claim of no production behavior change until the separate cutover.

### Advisory

This review is advisory; the `req` listener gap is the one issue I'd fix before merge given it sits on the hot path for nearly all proxied traffic — small fix, real blast radius.

— Independent Claude review

---

adamtasteslikegood:
Pushed a fix for the `req` listener gap flagged above: 8b3b3b2.

`server/proxy.ts` now attaches a no-op `req.on('error', () => {})` before `await getFlaskAuthHeader()`, so the request stream is never listener-less during the mint window — matching the pre-PR behavior where `req.pipe()` ran synchronously in the same tick.

Added a regression test (`server/proxy-auth.test.ts`) that uses a real `EventEmitter` (not a mock) to hold `forward()` suspended mid-`await` and asserts `req.emit('error', ...)` doesn't throw — this reproduces the crash pre-fix and passes post-fix. Also backfilled `on: vi.fn()` on the other `createFlaskProxy` req mocks in `proxy-auth.test.ts`/`server.test.ts` that didn't have it, since they'd otherwise break on the new `req.on(...)` call.

Verified locally: `type-check`, `lint` (`--max-warnings 0`), `prettier --check`, and the full `vitest` suite (259/259) all pass.

---

server/proxy.ts
```diff
@@ -31,11 +32,46 @@ export function createFlaskProxy(label = 'Flask') {
...
+    const authHeader = await getFlaskAuthHeader();
+    // The await above is a window the synchronous version did not have: a
+    // client can disconnect during it. Dispatching anyway would open a socket
+    // to Flask whose response has nowhere to go.
+    //
+    // Test only `res`. req.destroyed is ALSO true for every AI-endpoint request
+    // here — validation.ts fully consumes that body stream to buffer rawBody,
+    // which ends and destroys the IncomingMessage — so including it drops
+    // exactly the requests this proxy is most careful about replaying.
+    if (res.destroyed) return;
```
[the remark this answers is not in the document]
adamtasteslikegood: Someone is next in this file, so: test added in 396bb79, exactly per the suggested oracle.

The test holds the mint open with a deferred promise (so `forward()` is genuinely suspended at `await getFlaskAuthHeader()`), flips `res.destroyed` inside that window, releases the mint, and asserts `transport.request` was never invoked — plus `req.pipe` as a belt. The previously uncovered `if (res.destroyed) return` branch is now pinned.

Verified: type-check, lint (`--max-warnings 0`), prettier, and the full suite (260/260) all pass.

---

server/flask-auth.ts
```diff
@@ -0,0 +1,118 @@
...
+ * request: while the invoker check is disabled, Flask accepts the call and the
+ * site keeps working; once it is enabled, Cloud Run answers 403 and the
+ * throttled error below is the breadcrumb explaining why.
+ */
+export async function getFlaskAuthHeader(): Promise<string | null> {
+  const audience = getFlaskAudience();
+  if (!audience) return null;
+
+  try {
+    if (!cached || cached.audience !== audience) {
```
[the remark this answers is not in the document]
adamtasteslikegood: Verified against the code — the reset is non-atomic, but no interleaving produces a wrong outcome, so declining a change here. Walking the scenarios:

1. **Reset while another caller awaits the same promise.** Every caller reads `cached.client` into a local `client` before awaiting, so `cached = null` in one caller's catch cannot yank the promise out from under an in-flight awaiter. If the shared promise rejects, each awaiter lands in its **own** catch: the reset is idempotent, the log is throttled, and each returns `null` — which is this function's documented fail-open contract, not a new failure mode.

2. **Reset discarding a healthy client.** If caller A fails at `getRequestHeaders()` while caller B succeeded on the same client, A's reset throws away a working client and the next request rebuilds it — one extra `getIdTokenClient()` metadata round-trip, bounded and self-healing. Duplicate work, not incorrectness, and only on the request immediately following a mint failure.

3. **Concurrent double-construction** (the adjacent concern in the earlier `flask-auth.ts:84` thread): can't happen on the cold-start path — the check-and-assign has no `await` between `if (!cached …)` and `cached = {…}`, and JS runs that section to completion before any other call enters. Rebuild-after-failure (scenario 2) is the only path to a second construction, which also corrects the earlier thread's premise that two first-ever calls race.

A guarded reset (`if (cached?.client === myClient) cached = null`) would narrow scenario 2's window, but it adds bookkeeping to the one module whose contract is "never throws, worst case null" without changing any observable behavior — and the comment's own Details section concedes the race is "mostly harmless". Simplicity wins; no change.

---

docs/security/KAN-170_PUBLIC_EGRESS_REMEDIATION.md
adamtasteslikegood: No ticket existed — searched KAN for invoker/ingress/egress follow-ups and found only KAN-170/KAN-171 themselves. Filed now:

**[KAN-172](https://tasteslikegood.atlassian.net/browse/KAN-172)** — *Record express-frontend's security posture (invoker-iam-disabled + ingress-only guard) as a deliberate decision — KAN-170 twin.* Linked to KAN-170, carries your framing verbatim: ingress is the differentiator, not IAM; the posture is acceptable but must be a recorded decision rather than an accident of DRS. "Done" includes linking the runbook bullet back to the ticket, so the twin defect no longer lives only in a runbook paragraph.

The drift-detection half (annotation reappearing, ingress widening later) is deliberately split into [KAN-173](https://tasteslikegood.atlassian.net/browse/KAN-173) — recording and detecting are different work with different owners' cadence.

---

scripts/gcloud/kan170_verify.sh
adamtasteslikegood: Agreed on all points, and the follow-up now exists: **[KAN-173](https://tasteslikegood.atlassian.net/browse/KAN-173)** — *Detect KAN-170 recurrence* (linked to KAN-170).

Scope captured from this thread plus the adjacent ones:
- Scheduled check asserting `invoker-iam-disabled` is **absent across all Cloud Run services** (allowlisting express-frontend only while the KAN-172 decision stands), and that express-frontend's ingress is still `internal-and-cloud-load-balancing` — covering the re-disable, the later widening, and the future-service-with-DRS-default cases you listed.
- The KAN-138 precedent is cited in the ticket as the pattern to copy, including the second half: wiring it to a schedule **and** an alert path a human actually sees, not a red run nobody watches.
- Folded in the post-cutover observability gap from the `flask-auth.ts:95` thread — a log-based alert on `[FlaskAuth]` mint failures — so that promise now has a tracked home too.

Kept out of this PR per your sequencing point; nothing here blocks the merge.

---

adamtasteslikegood:
## Security review — no HIGH/MEDIUM findings

Focused security review of this PR's full diff (10 files, +1318/−1), run at head `8b3b3b2` against `dev` with full repo context (Express security middleware, Flask's `Authorization`-consuming paths, the deploy pipeline). Method: an independent finder pass over the diff and the PR-head sources, followed by verification of every load-bearing claim against the implementation. The later `396bb79` is test-only and does not change the analysis.

**Result: no HIGH- or MEDIUM-severity vulnerability is newly introduced.** Attack surfaces examined and why each is clean:

- **Header spoofing/smuggling** (`server/proxy.ts`): `delete headers['x-serverless-authorization']` runs unconditionally after the `...req.headers` spread and before the conditional set of the real token. Node lowercases incoming header names, so casing variants and duplicates collapse onto the deleted key; in the mint-failure path nothing re-adds the header. A forged client token cannot reach Cloud Run or Flask — all four cases pinned by `server/proxy-auth.test.ts`.
- **Token exfiltration**: the audience (`resolveFlaskAudience`) and the proxy destination both derive solely from the trusted `FLASK_BACKEND_URL` env var — no request-controlled input influences either. Tokens are minted only for `https://` `*.run.app` origins and sent only to that origin over TLS; no certificate-validation bypass anywhere in the diff.
- **Fail-open mint path**: not an authentication bypass. Pre-cutover there is no check to bypass; post-cutover a token-less request is rejected by Cloud Run's invoker IAM check at the edge — the security outcome is fail-closed, with availability as the only cost.
- **Authorization-boundary preservation**: the token travels in `X-Serverless-Authorization` (consumed and stripped by Cloud Run), leaving the client's `Authorization` intact for Flask's `require_admin` and `require_pubsub_oidc` — no privilege confusion between proxy identity and end-user credentials.
- **Secret exposure in logs**: the minted token is never logged. The `flask-auth.ts` failure log emits only audience + error message (throttled); the proxy's dispatch-failure log emits sanitized method/URL/message.
- **Operator scripts** (`scripts/gcloud/kan170_*.sh`): inputs are env vars and gcloud API output, empty-value guards are present, mutation requires explicit `--apply`, no `allUsers`/public IAM grants. The `roles/run.invoker` grant landing on the shared default compute SA is a scoping tradeoff the runbook documents as accepted and separately tracked — not an exploitable vulnerability.

The PR does what it claims: it adds a proxy-identity credential without changing live authorization behavior, and defers the enforcement flip to a separate post-verification release.

---

PR #3321, opened 2026-07-31
docs+ci(release): lock the train procedure, gate the stations, add an interactive driver [KAN-191]

adamtasteslikegood:
Closes the loop opened by KAN-191: the release train's procedure now lives in the repo, the station that catches pointer drift now gates a merge, and a driver holds the ten-step sequence so a human does not have to.

Three commits, in the order Adam asked for: **docs → gate → runner**.

## 1. Docs — `scripts/release/RUNBOOK.md`

The ordered procedure lived on **KAN-138**, where a checkout cannot read it. Predictably it drifted from `CLAUDE.md`, and the copy in the repo was the wrong one.

Ten numbered steps, each with its precondition, command, and how you know it worked. Four traps written in **at the step where they bite**, each because it cost a real release:

| Trap | Evidence |
| --- | --- |
| A Backend promotion goes stale the moment anything else lands on Backend `dev` | v0.4.8: #258 promoted 19:17, #259 merged 03:43. The step list said "done"; `main` did not have the code. Cost promotion #261. |
| `git submodule update --remote Backend` can never satisfy Station 3 | `.gitmodules` tracks `dev`; Station 3 wants `main`. See below. |
| A spent version deploys nothing and reports success | `release.yml` skips the tag, the Release, and the trigger — green PR, green workflow, no deploy, no error. |
| Production verification must grep every served chunk | v0.4.8 was live for ~20 min while a `main-*.js`-only poller said "not deployed". |

### The CLAUDE.md fix

`CLAUDE.md` step 2 said `git submodule update --remote Backend`. That resolves to Backend **`dev`**'s tip; `train-verify.sh` Station 3 requires **`main`**'s tip. **These can never agree**, because each promotion + back-sync cycle produces two distinct merge commits:

```
main = merge(old_main, dev_tip)     7b6347e = merge(d48f06a, 73fb091)
dev  = merge(dev_tip,  main)        3fac56b = merge(73fb091, 7b6347e)
```

So `dev`'s tip is structurally never `main`'s tip again. Following the documented instruction produced a Station 3 block **every release** — v0.3.9 shipped that way, v0.4.8 nearly did (fixed by #3320). Replaced with the correct command plus the reason, in both places CLAUDE.md mentioned it.

## 2. Gate — `train-verify` now blocks main-targeting PRs

`train-verify.sh` has been correct since KAN-138 and **gated nothing**: `release-train.yml` runs it on `workflow_dispatch` and a daily cron, so no merge ever depended on it. v0.4.8's bad pointer was corrected because a human read the output.

New `release-train` job in `pr-gate.yml`, `if: github.base_ref == 'main'`, wired into the `gate` aggregator's `needs`. `dev`-targeting PRs skip it deliberately — a lagging pointer and unpaid back-sync are normal mid-cycle, and blocking them teaches people to ignore the check. `gate` already treats `skipped` as passing.

**The plumbing is proven, not assumed.** It is copied verbatim from `release-train.yml`, and scheduled run [30557811029](https://github.com/adamtasteslikegood/tasteslikegoodtheangularsvegancookbook/actions/runs/30557811029) exited 1 with `BLOCK Backend main→dev back-sync owed` — meaning that checkout+fetch pair resolves refs in **both** repos and fails for the reason the check exists, rather than dying at exit 2 on its own broken plumbing. That distinction is the one this repo keeps relearning.

## 3. Runner — `scripts/release/train-run.sh`

Stops before every mutating step, prints the state it is about to act on, and requires a **typed word** (`PROMOTE` / `BACKSYNC` / `RELEASE`) rather than y/n. A checklist persists in `.agent-work/` (gitignored), so an interrupted train resumes instead of restarting from memory.

`--status` and `--verify-only` are read-only, `--dry-run` runs nothing, and a **non-TTY refuses to confirm** rather than proceeding — verified.

### Two bugs found by running it, not reasoning about it

1. **False "promotion owed" against a just-released repo.** After a back-sync, Backend `dev` is *permanently* 1 ahead of `main` with an identical tree — the steady state, not unshipped work. Counting commits reports a promotion owed forever, which trains the reader to ignore the line. Now discriminates by tree, as `train-verify` does.
2. **`grep -c || echo 0` produced the two-line string `0\n0`** — `grep -c` already prints `0` and exits 1 — which blew up `$(( ))` with an arithmetic syntax error and derailed the entire verify mode.

### Step 9 verified in both directions against the live v0.4.8 deploy

```
POSITIVE  marker OWNERSHIP_ORPHANED_GUEST_ROW
  chunk-KLTA5W4T.js  257687 bytes  hits=1
  chunk-L4HQU7XA.js   41135 bytes  hits=1
  main-QIIFTY6Y.js   560638 bytes  hits=0      <-- the whole point
  ✓ marker found in 2 place(s) — this release IS live        exit 0

NEGATIVE  marker ZZZ_NOT_IN_ANY_BUILD_ZZZ
  ✗ marker not found in any served asset                     exit 1
```

`main-*.js` has **zero** hits while the chunks have them — exactly the shape that made a main-only poller report "not deployed" while v0.4.8 was serving. And a marker no build contains fails, so the check can fail for the reason it exists.

## Verification

- `bash -n` clean; exit codes checked: walk `1` (real back-sync debt), `--status` `0`, bad arg `2`, negative verify `1`.
- YAML parsed: `release-train` present, `if: github.base_ref == 'main'`, in `gate.needs` (11 entries).
- `train-verify.sh --for-release` exits 1 locally right now for three real reasons — the script fails, it is not a no-op.
- `format:check` clean; checklist state confirmed gitignored (`.gitignore:79`).

## Still owed on KAN-191

**A red test PR against `main` proving this job blocks the merge.** A green run proves the job exists, not that it can fail — that is the acceptance criterion on the ticket and it is not satisfied by this PR. Doing it here would mean opening a throwaway `dev → main` PR, which is a release-shaped action; better as its own deliberate step.

Related follow-ups filed from the v0.4.8 review: KAN-189, KAN-190, KAN-192.

https://claude.ai/code/session_01RJTYg2F4r5Bcy6aX6hW8Ga

---

### DIFF
--- a/.github/workflows/pr-gate.yml
+++ b/.github/workflows/pr-gate.yml
@@ -176,6 +176,48 @@ jobs:
       - uses: actions/checkout@v7
       - run: bash scripts/gcloud/kan170_verify.sh --self-test
 
+  # Release-train stations, on main-targeting PRs ONLY (KAN-191).
+  #
+  # train-verify.sh has existed since KAN-138 and correctly exits 1 on blocking
+  # drift — but release-train.yml runs it on workflow_dispatch and a daily cron,
+  # so nothing ever merged on its result. v0.4.8 shipped a pointer that Station 3
+  # blocked; it was corrected because a human read the output, not because CI
+  # stopped the merge. This job is what makes the station a gate.
+  #
+  # dev-targeting PRs skip it deliberately: a lagging pointer and an unpaid
+  # back-sync are normal mid-cycle states, and blocking them would train people
+  # to ignore this check. `gate` treats `skipped` as passing.
+  release-train:
+    name: Release train — verify stations
+    if: github.base_ref == 'main'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          submodules: true
+          fetch-depth: 0
+      # checkout fetches only the PR ref; the stations compare origin/dev against
+      # origin/main in BOTH repos, so those remote-tracking refs must exist.
+      # Without this the script dies at exit 2 ("could not determine") rather
+      # than passing — the right failure, but a useless one.
+      - name: Fetch full refs for both repos
+        run: |
+          git config --unset-all remote.origin.fetch || true
+          git config --add remote.origin.fetch '+refs/heads/*:refs/remotes/origin/*'
+          git fetch origin --prune --tags
+          git -C Backend fetch origin --prune
+      # --no-fetch because the step above already did it; --for-release still
+      # queries origin for the tag regardless, which is the one lookup that must
+      # not be guessed.
+      #
+      # This exact checkout+fetch pair is proven, not assumed: release-train.yml
+      # runs it daily, and run 30557811029 (2026-07-30) exited 1 with
+      # "BLOCK Backend main→dev back-sync owed" — i.e. it resolved refs in BOTH
+      # repos and failed for the reason the check exists, rather than dying at
+      # exit 2 on its own plumbing.
+      - name: Verify release-train stations
+        run: ./scripts/release/train-verify.sh --for-release --no-fetch
+
   # ── Gate ──────────────────────────────────────────────────────────────────
   # Single required check for branch protection. Fails if ANY job above fails.
 
@@ -192,6 +234,7 @@ jobs:
       - changelog-check
       - canonical-recipes
       - posture-check-self-test
+      - release-train
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -163,12 +163,23 @@ Never commit directly to `main` or `dev`. Always branch off `dev`.
 
 Since 2026-07-18 this is enforced by branch protection: `dev` and `main` both carry required status checks `Gate — all checks passed` (the pr-gate aggregate, which includes the Express Docker image build), `Analyze (javascript-typescript)`, and `Dependency Review` (strict off, 0 approvals, admin break-glass audited; SPEC-01 §4.3, `docs/ci/refresh/`).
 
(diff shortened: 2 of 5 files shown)


adamtasteslikegood:
## Review — automated (scheduled routine, triggered on PR open)

**Overview:** docs+CI change delivering the KAN-191 loop closure: `RUNBOOK.md` centralizes the ordered release procedure (previously stranded on an unreachable Jira ticket, KAN-138), `pr-gate.yml` gains a `release-train` job that actually gates `main`-targeting PRs on `train-verify.sh --for-release` (previously only advisory via cron/dispatch), and `train-run.sh` is a new interactive driver with a persisted, gitignored checklist.

**Quality:** high. Two real bugs are documented as caught by *running* the tooling rather than by inspection — the commit-count-based "promotion owed" false positive (fixed by comparing trees instead) and the `grep -c || echo 0` arithmetic bug in the production-verification step. The new CI job's checkout+fetch plumbing is validated against an actual `release-train.yml` run (30557811029) that failed for the intended reason rather than dying on broken ref resolution. `CLAUDE.md`, `scripts/release/README.md`, and the new `RUNBOOK.md` agree with each other on the corrected `git -C Backend checkout origin/main` (not `--remote`) instruction. CI is green across the board.

**One suggestion (non-blocking):** the auto-generated PR titles in `train-run.sh` (the Backend promotion PR and the cookbook release PR) hardcode a literal `[KAN-###]` placeholder rather than a real ticket number. Per `CLAUDE.md`'s PR lifecycle section, Jira's GitHub integration links a PR by scanning the *title* for a real key (`KAN-123`) — `KAN-###` won't match, so every driver-opened PR needs a manual title edit before Jira picks it up. Since `--marker` already shows the pattern for collecting a value interactively (step 9), the same approach (or just an env var / flag) would let `PROMOTE`/`RELEASE` confirmations produce a title that's actually correct, rather than reliably wrong until hand-fixed.

No correctness or safety issues found; nothing here blocks merge.

---

PR #3336, opened 2026-08-02
feat(ci): make the sprint-lane check a blocking gate [KAN-200]

adamtasteslikegood:
Sprint 4 retrospective action → Sprint 5 **S8** (KAN-200 ↔ RCP-70), committed on Adam's call. Retro success measure, verbatim: *"a PR orphaning a `sprint-N` row fails CI."*

Until now `check_sprint_lane.sh` was a script run on demand — the same distinction that made the Alembic head-check look like a gate for weeks while only running on release-train dispatch.

## ⚠️ This PR cannot go green until two secrets exist

`ATLASSIAN_EMAIL` and `ATLASSIAN_API_TOKEN` are **not** GitHub repository secrets today (verified with `gh secret list`). Without them the job exits 2 and fails — **deliberately**. A gate that silently passes when it cannot run is precisely the failure mode this work exists to correct, so I did not add a "skip if no secret" escape hatch.

**This is a decision, not a chore:** it puts an Atlassian API token in CI, readable by any workflow in the repo. Your call. I can set them from your local `.env` with `gh secret set` if you want me to.

## What changed

**Wiring** — new `sprint-lane` job in `pr-gate.yml`, added to `gate.needs`. `gate` is the required status check, so **branch protection needs no change**.

**Script** — two fixes:
- `.env` is now optional; CI injects the credentials as env vars instead.
- Credential absence now exits **2** explicitly. It previously used `${VAR:?...}`, which aborts with status **1** — and this script's own contract defines 1 as *"an orphan was found, the lane is drifting."* A CI job with no credentials would have been reported as real lane drift. Pre-existing, but it only starts mattering once CI branches on the exit code.

## Proved failing first

A gate that already passes proves nothing, so all three paths were exercised:

| Condition | Result |
|---|---|
| no credentials | `exit 2` — `FAIL(2): ATLASSIAN_EMAIL unset` |
| **orphaned sprint row** | `exit 1` — `ORPHAN KAN-191`, *"the lane is drifting"* |
| clean lane | `exit 0` — `8 open KAN row(s), 0 orphan(s)` |

The orphan case was produced by transiently labelling KAN-191 `sprint-1` — a label sorting below `sprint-5`, so the script's default newest-label pick was unaffected — and restored in a `finally` block. Labels verified back to `['ci','documentation-drift','followup-v0.4.8','release-train']`.

Incidental finding: **Jira's JQL index lags ~5s behind a label write.** The first orphan test reported "no open KAN rows" because it queried before the index caught up.

## Two limits survive the wiring

Documented in both the script and the workflow rather than left to be discovered:

1. **It passes vacuously** when no row carries the label — it detects lane *drift*, not a missing sprint.
2. **Fork PRs (incl. Dependabot) are skipped** — no secrets there — and `gate` counts skipped as passing. Same carve-out `release-train` already documents. A fork PR is not lane-checked.

## Docs

**`CLAUDE.md`** — new session-start subsection: *read the Confluence retro before chartering*. A sprint retro lives in two places with different content, and only the repo close-out is visible from a checkout. That is exactly how Sprint 5's charter missed two retro actions, this one included.

**`specs/SPRINT_5_PLAN.md`** — S8 recorded; WIP 7 → 8 (**4× by count, ~1.5× by rate**); R2's mitigation moved from aspirational to committed; and RCP-55 clarified — its **status** is Done, its **title** is still wrong, and those are different things. The retitle stays yours per Sprint 4's close-out.

https://claude.ai/code/session_01D9HBZQ4rRxotou198eaphu

---


adamtasteslikegood:
## Unblocked and green — the gate is live and proven on all three paths

Adam added `ATLASSIAN_EMAIL` and `ATLASSIAN_API_TOKEN` as repository **secrets** (verified they landed in the secrets namespace, not variables — this workflow reads `secrets.*`, so that distinction would have silently kept the job failing).

**The green path is now proven in real CI**, not just locally:

```
lane check: sprint-5 — 8 open KAN row(s), 0 orphan(s)
PASS: every open KAN row labelled sprint-5 is linked to an RCP row.
```

All three exit paths are now demonstrated end-to-end:

| Path | Where proven | Result |
|---|---|---|
| no credentials | **live CI**, run 30738190015 | `FAIL(2): ATLASSIAN_EMAIL unset` → `exit 2` → `Gate` failed → PR `BLOCKED` |
| orphaned row | local, transient `sprint-1` label on KAN-191 | `exit 1` — `ORPHAN KAN-191`, *"the lane is drifting"* |
| clean lane | **live CI**, after secrets | `exit 0` — 8 rows, 0 orphans |

The middle row is the one that matters: the retro's success measure was *"a PR orphaning a `sprint-N` row fails CI"*, and the failing-credentials run independently confirmed the other half of that chain — `sprint-lane` failure genuinely propagates through `gate.needs` to block the merge. Before this, nothing in CI could do that.

Required checks: **Gate — all checks passed**, **Analyze (javascript-typescript)**, **Dependency Review**, **Sprint lane — KAN/RCP link** — all pass, zero failures.

## Also settled here: RCP-55

Adam approved the retitle, closing the last open row of the Sprint 4 retro's actions table.

> **old:** Publishing succeeds regardless of which account or guest session owns the row
> **new:** Publishing a recipe owned by another account or guest session is refused with a distinct, actionable reason — never reported as success (INV-4)

The old title asserted the flattering outcome — publish always succeeds — which would have required weakening `same_owner`, i.e. Sprint 4's dominant risk **R7**. v0.4.8 shipped the opposite: the refusal preserved and made *legible*. Status was already `Done` and that was correct; closing a row doesn't fix what it claims.

Previous title recorded verbatim in a comment on the ticket rather than erased — Sprint 4's close-out was right that this move requires a human, and the rule held.

**The Sprint 4 retro's actions table is now fully accounted for:** three delivered into Sprint 5 scope (S5/S3/S8), one done at charter time, one settled directly.

adamtasteslikegood:
## Review round 3 — Copilot right again, and this one would have broken production CI

**Finding 1 (Dependabot): correct, and serious.** Verified rather than reasoned about:

```
$ gh pr list --author app/dependabot
PR 3335 | head owner: adamtasteslikegood | branch: dependabot/npm_and_yarn/...
```

Dependabot PRs are **not forks** — they're same-repo branches, so `head.repo.full_name == github.repository` evaluates **true** for them, while Actions secrets are still withheld. Net effect of my original condition: the job would have **run on every Dependabot PR, exited 2, failed `gate`, and blocked every dependency update**. #3335 is open right now and would have broken the moment this merged.

My workflow comment asserted Dependabot was skipped. It said so because I assumed it, not because I checked — the same "inherited assertion never verified against the thing it describes" pattern Copilot has now caught on this branch three times running.

Adam had already pushed the same fix in 4dc2b0f. I rebased onto it rather than clobbering it, and **kept both tests**:

```yaml
if: >-
  github.event.pull_request.head.repo.full_name == github.repository
  && github.actor != 'dependabot[bot]'
  && github.event.pull_request.user.login != 'dependabot[bot]'
```

Not redundancy for its own sake: `github.actor` is the actor of the *triggering event*, so on a **manual re-run** it becomes the human who pressed the button and a re-run of a Dependabot PR can slip past that test alone. `pull_request.user.login` is the PR author and never changes. Evaluated against all four shapes — same-repo → RUNS; Dependabot → skipped; **Dependabot + human re-run → skipped**; fork → skipped.

**Finding 2 (scope overclaim): also correct.** The job invokes `check_sprint_lane.sh` with no label argument, so it asserts only the **newest** `sprint-N` label — the current sprint. An orphan on an older superseded label does not fail CI.

That is the intended scope (a closed sprint's rows are `Done` and filtered out anyway), but the retro's phrasing *"a PR orphaning a `sprint-N` row fails CI"* is broader than what this enforces, and I'd copied the broad phrasing into the workflow comment. The narrower truth is now written in both the workflow and the script header. The script's "two limits" is now **three**, and no longer conflates forks with Dependabot — they're different cases needing different tests.

**Tally on this branch: Copilot 3, me 0.** Every one was an assertion in a comment that didn't match the code beside it.

adamtasteslikegood:
## Review round 4 — premise incorrect, change made anyway

**Finding:** an inline code span in `specs/SPRINT_5_PLAN.md` split across a newline, said to break in GitHub-flavored Markdown.

**Verified against the source, and the premise doesn't hold.** CommonMark treats a line ending inside a code span as a space, so this:

```
`grep -rn check_sprint_lane
.github/workflows/`
```

rendered as one correct span — `grep -rn check_sprint_lane .github/workflows/`. Confirmed by matching the span across the newline in the raw file. Nothing was broken on the rendered page.

**Reworded regardless, for a different reason than the one given.** The span sat exactly on Prettier's wrap column, so every reformat re-split it — and a command broken mid-line is harder to read and to copy out of the raw file even when it renders correctly. The sentence is restructured so the span can't land on the wrap.

Worth distinguishing from rounds 1–3: those were **substantive and correct** — a false `delivery_loop_gate.py` claim, an impossible date range, a missing KAN-157 disposition, and a condition that would have blocked every Dependabot PR. This one is cosmetic and its stated reason is wrong. Fixed on its merits, not on the reasoning.

Running tally on this branch: **3 real findings, 1 cosmetic-with-a-wrong-premise.** The three real ones were each an assertion in a comment that didn't match the code beside it, which is a pattern worth me carrying forward rather than Copilot having to catch a fourth time.

PR #3344, opened 2026-08-06
chore(backend): bump submodule for the guest-merge dedup fix [KAN-186]

adamtasteslikegood:
## What

Bumps the `Backend` submodule pointer from `7b6347e` to **`fd2d1c1`** (Backend `dev` tip).

Backend [#267](https://github.com/adamtasteslikegood/tasteslikegood.com/pull/267) merged; a Backend change cannot reach production without this pointer, so this is the second half of that ship.

## What it picks up

| Commit | Change |
| --- | --- |
| `fd2d1c1` (#267) | **KAN-186** — guest→login merge runs the duplicate-recipe check instead of a bulk `UPDATE`. Sprint 5 item **S1**. |
| `bb7e167` (#266) | **KAN-183** — Backend half of the `synchronize` review trigger (already Done). |
| `5806a58`, `e5e07a0`, `50ce1bc` | Dependency bumps (claude-code-action, ddtrace, uv base image). |

## KAN-186 in one line

Logging in after saving a recipe as a guest no longer silently adds a second copy of a recipe the account already owns. Publish-time confirmation (INV-3) was catching this only *after* the duplicate row existed.

Backend-side verification on #267: 9 new tests, the 4 behaviour tests confirmed failing against the previous implementation first; full suite 357 passed; black/flake8/mypy/pip-audit/CodeQL/Docker green. Both reviews (Copilot, independent Claude) returned no requested changes.

## Migration / release safety

- **No migration in the range** — `git diff 7b6347e..fd2d1c1 -- migrations/` is empty, and Backend is on a single Alembic head. Risk R8 does not apply; this drags no release.
- **This is the routine dev-tracking bump, not a release pin.** The release pointer must be Backend **`main`**'s own SHA (`train-verify.sh` Station 3) — re-pin at release time rather than shipping this SHA as the release pointer. Pinning `dev`'s tip as a release pointer is the trap that blocked v0.3.9 and nearly v0.4.8 (KAN-191).

_Authored by Claude on Adam's behalf_

---


adamtasteslikegood:
**Fixed in `0cf55bd`** — the pointer now pins `fd2d1c1`, the merge commit that landed Backend #267 on `dev`.

Both reviews raised the same discrepancy (Copilot inline on `Backend:1`, and observation 1 of the independent Claude review): the body said `fd2d1c1`, the gitlink said `edbc1fe`. That was real, and the reviewers picked the right target.

Verified before re-pinning:

```
$ git -C Backend merge-base --is-ancestor fd2d1c1 origin/dev && echo YES
YES ancestor of origin/dev

$ git -C Backend rev-parse fd2d1c1^{tree} edbc1fe^{tree}
d2d16e4994ece94dc4eac5716e9c3cbd378c8344
d2d16e4994ece94dc4eac5716e9c3cbd378c8344   # identical
```

So this is a zero-behaviour-change commit: same tree, same build. What it buys is that *"which Backend SHA carried KAN-186 into the cookbook?"* is answerable from a commit on `dev`'s first-parent history, rather than one reachable only from `fix/kan-186-guest-merge-dedup` — a branch that gets deleted. That is exactly the auditability argument observation 1 made.

---

**On observation 2 (dev-only SHA) — agreed, and no change here is correct.**

`fd2d1c1` is reachable from Backend `origin/dev` but not `origin/main` (`main` is still at `7b6347e`). That is right for a `dev`-lane pointer bump and wrong for a release pointer. Recording the follow-through explicitly so it is not lost, since a release is now in flight this sprint:

before the cookbook `dev → main` release PR, Backend must be promoted `dev → main`, and the pointer re-pinned to **Backend `main`'s own SHA** via `git -C Backend checkout origin/main` — never `git submodule update --remote`, which resolves to `dev` because `.gitmodules` tracks it. `train-verify.sh` Station 3 blocks on precisely this, and it is the KAN-191 trap that shipped v0.3.9 wrong and nearly shipped v0.4.8 wrong.

PR #3361, opened 2026-08-08
fix(release): --bump must read the pointer this tree ships, not origin/dev's [KAN-138]

adamtasteslikegood:
Found on the guard's **first real use**, cutting v0.4.9. This is a follow-up to #3359.

## The bug

`do_bump` compared `gather()`'s `$POINTER` — `git rev-parse origin/dev:Backend` — against Backend `main`. That is correct for the status display and **wrong at step 5**: RUNBOOK step 4 stages the re-pin *locally on the release branch*, and it does not reach `origin/dev` until the release PR merges.

So immediately after step 4 pinned Backend `main` exactly as instructed, the guard read `dev`'s **old** pointer and refused:

```
· version   0.4.8 → 0.4.9
✗ pointer   fd2d1c1cfd58 does NOT pin Backend main (f1219e81be02)
error: refusing to scaffold a CHANGELOG against a non-main pointer
```

`fd2d1c1` is what `origin/dev` still pins. `f1219e8` is what the release branch had already staged and would ship.

## The worse half

The same stale value was passed to the CHANGELOG scaffold:

```bash
python3 - "$to" "$POINTER_SHORT"   # ← origin/dev's pointer
```

Had the comparison happened to pass, the generated section would have **named the old Backend SHA** — the exact wrong-SHA outcome this guard exists to prevent, and the KAN-191 trap wearing the guard's own uniform. A false refusal is noisy; this one would have been silent and wrong.

## Fix

Read the **index** (`git rev-parse :Backend`) — it holds the gitlink that will actually be committed — falling back to `HEAD`. `gather()`'s `$POINTER` is left alone, since the status display genuinely wants what `origin/dev` ships today.

## Verified both directions, not assumed

```
$ git rev-parse :Backend            # f1219e81be02 (step 4 staged)
$ ./scripts/release/train-run.sh --dry-run --bump 0.4.9 ; echo $?
· pointer f1219e81be02 (Backend main) ✓
0

$ git restore --staged Backend      # index falls back to HEAD's old value
$ ./scripts/release/train-run.sh --dry-run --bump 0.4.9 ; echo $?
✗ pointer fd2d1c1cfd58 does NOT pin Backend main (f1219e81be02)
2
```

Passes when the pointer is right, **still fails when it is wrong**. A guard that only ever passes is not a gate — so the negative case was re-run explicitly after the change rather than reasoned about.

## Why this took a real release to find

Every check in #3359 ran against a working tree where step 4 had *not* been performed, so `origin/dev`'s pointer and the local pointer were the same value and the bug was invisible. It only separates during an actual cut, between step 4 and step 5 — a window that exists exactly once per release.

## Jira

[KAN-138](https://tasteslikegood.atlassian.net/browse/KAN-138)

https://claude.ai/code/session_012EfffNRnHWdCc7Pn4rEfMp

[KAN-138]: https://tasteslikegood.atlassian.net/browse/KAN-138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

### DIFF
--- a/scripts/release/train-run.sh
+++ b/scripts/release/train-run.sh
@@ -429,14 +429,30 @@ do_bump() {
 
   info "version   $VERSION → $to"
 
+  # Read the pointer THIS WORKING TREE will ship, not origin/dev's.
+  #
+  # gather()'s $POINTER is `git rev-parse origin/dev:Backend`, which is right
+  # for the status display but wrong here: RUNBOOK step 4 stages the re-pin
+  # locally on the release branch, and it does not reach origin/dev until this
+  # release PR merges. Comparing against origin/dev at step 5 therefore refuses
+  # a correctly-ordered release — caught on this guard's first real use, cutting
+  # v0.4.9, where step 4 had just pinned Backend main and the guard still saw
+  # dev's old pointer.
+  #
+  # The index is the right source: it holds the gitlink that will be committed.
+  local local_pointer local_short
+  local_pointer=$(git rev-parse :Backend 2>/dev/null || git rev-parse HEAD:Backend 2>/dev/null || echo "")
+  [ -n "$local_pointer" ] || die "could not read the Backend gitlink from this tree"
+  local_short=${local_pointer:0:12}
+
   # Step 5 comes AFTER step 4 for a reason: the CHANGELOG section names the
   # pinned SHA, and train-verify checks that name against the actual gitlink.
   # Scaffolding it while the pointer still pins Backend dev bakes the KAN-191
   # trap into the prose, where it reads as deliberate.
-  if [ "$POINTER" = "$BE_MAIN_FULL" ]; then
-    info "pointer   $POINTER_SHORT (Backend main) ✓"
+  if [ "$local_pointer" = "$BE_MAIN_FULL" ]; then
+    info "pointer   $local_short (Backend main) ✓"
   else
-    bad "pointer   $POINTER_SHORT does NOT pin Backend main (${BE_MAIN_FULL:0:12})"
+    bad "pointer   $local_short does NOT pin Backend main (${BE_MAIN_FULL:0:12})"
     info "RUNBOOK step 4 comes first, or this CHANGELOG will name the wrong SHA:"
     info "  git -C Backend fetch origin --prune && git -C Backend checkout origin/main && git add Backend"
     die "refusing to scaffold a CHANGELOG against a non-main pointer"
@@ -447,7 +463,7 @@ do_bump() {
     return 0
   fi
 
-  python3 - "$to" "$POINTER_SHORT" <<'PY' || die "bump failed"
+  python3 - "$to" "$local_short" <<'PY' || die "bump failed"
 import json, re, sys, datetime
 
 to, pointer = sys.argv[1], sys.argv[2]


adamtasteslikegood:
**Correction — this PR contained a file its description did not mention.**

Merge commit `8b000ba` touched **two** files, not one:

```
 Backend                      |  2 +-
 scripts/release/train-run.sh | 24 ++++++++++++++++++++----
```

The `Backend` submodule pointer moved `fd2d1c1` → `f1219e8` here. That was not deliberate and was not in the PR body.

**Cause.** While cutting v0.4.9 I had already performed RUNBOOK step 4 on the release branch — `git -C Backend checkout origin/main && git add Backend` — so the re-pin was sitting **staged in the index**. Creating this branch carried the staged entry across the switch, and `git add scripts/release/train-run.sh && git commit` then committed *the whole index*, not just the path I had most recently added. `git status` would have shown `M  Backend`; I filtered it out of the output I was reading.

**Impact: benign, but worth stating precisely rather than waving through.**

- `f1219e8` is Backend `main`'s own tip (the merge commit of Backend #271), which is exactly where the release needs the pointer. `train-verify` reports `matches-backend-main` *and* `matches-backend-main-content`.
- Backend `main` had already been promoted and back-synced at this point, so no unreviewed Backend code rode along — `f1219e8`'s tree is the promotion of commits that went through Backend #267/#271.
- The effect was to land step 4 on `dev` one PR earlier than intended. The v0.4.9 release commit is consequently version + CHANGELOG only, with the pointer already correct at `HEAD`.

**What it could have been.** Had the staged pointer been anything else — Backend `dev`'s tip, or a stale SHA — this PR would have quietly shipped a wrong submodule pointer to `dev` under a title about a shell script, and the next release would have inherited it. That is the KAN-191 failure mode arriving through an unrelated PR, which no gate here would have caught: `train-verify` only blocks on `main`-targeting PRs, and this one targeted `dev`.

**Generalisable, and going into my working notes:** after `git add <path>`, `git commit` still commits everything already in the index. When a submodule re-pin is staged as part of a release cut, any commit made on any branch until it is committed or unstaged will absorb it. Read the full `git status` before committing during a release cut — do not filter it.

No revert: reverting would move `dev`'s pointer *away* from Backend `main` and break the release that is now mid-flight. Recording it instead.

scripts/release/train-run.sh
```diff
@@ -429,14 +429,30 @@ do_bump() {
...
+  # for the status display but wrong here: RUNBOOK step 4 stages the re-pin
+  # locally on the release branch, and it does not reach origin/dev until this
+  # release PR merges. Comparing against origin/dev at step 5 therefore refuses
+  # a correctly-ordered release — caught on this guard's first real use, cutting
+  # v0.4.9, where step 4 had just pinned Backend main and the guard still saw
+  # dev's old pointer.
+  #
+  # The index is the right source: it holds the gitlink that will be committed.
+  local local_pointer local_short
+  local_pointer=$(git rev-parse :Backend 2>/dev/null || git rev-parse HEAD:Backend 2>/dev/null || echo "")
```
adamtasteslikegood, line 444: **Checklist step 5 will misreport "not done" during the exact window this fix targets.**

`do_bump` now correctly writes the CHANGELOG naming `local_short` (the index's Backend pointer, e.g. `f1219e81be02`) instead of the stale `POINTER_SHORT`. But `changelog_state()` — used by `derive_steps()` for the step-5 checkbox, and by `print_checklist` — still compares the CHANGELOG body against the unchanged global `$POINTER` (`git rev-parse origin/dev:Backend`), which is exactly the stale value this PR proves is wrong during this window:

```bash
changelog_state() {
  python3 - "$VERSION" "$POINTER" <<'PY' 2>/dev/null || echo "missing"
...
    if pointer.startswith(sha) or sha.startswith(pointer[:7]):
```

Concretely, with this PR's own numbers: after a correct `--bump 0.4.9` (index/local pointer = `f1219e81be0280e687d3bd1f86d345e613475229`, matching Backend main), running the driver again in `--status`/walk mode before the release PR merges into `dev` still has `$POINTER` = the old `fd2d1c1cfd583fc234530042242be967699d8d90`. Neither `pointer.startswith(sha)` nor `sha.startswith(pointer[:7])` holds between `"fd2d1c1cfd58..."` and `"f1219e81be02"`, so `changelog_state()` returns `"section-only"` instead of `"named"`, and:

```bash
if [ "$(changelog_state)" = "named" ] && [ -z "$TAG_EXISTS" ]; then
    state_set 5 "done"
else
    state_set 5 ""
fi
```

sets step 5 back to unticked. So immediately after correctly completing step 5 with the fixed `--bump`, `--status` shows `[ ] 5. Version bump + CHANGELOG on a branch → cookbook dev` as incomplete — the "checklist that cannot be completed is one nobody reads" failure mode this file's own comments warn about, reintroduced by this fix for precisely the pre-merge window it was built to handle. `changelog_state()` should be fed the same local/index pointer `do_bump` now uses instead of the stale `$POINTER`.

---

adamtasteslikegood (commented):
Single-pass review (Agent tool unavailable this session — no multi-agent fan-out or subagent verify pass ran). 1 finding, traced concretely against this PR's own SHAs.

---

PR #3374, opened 2026-08-09
docs(sprint): Sprint 6 timebox, resolved counts, and the dedup query runbook [KAN-213]

adamtasteslikegood:
Sprint 6 is created and **active** — id 47, board 168, **2026-08-08 → 2026-08-13**, RCP-71/72/39/73 in the box.

## R2 retired before the box opened

The dominant *likelihood* risk was that prod DB access friction would stall the sprint exactly where KAN-182 predicts. It didn't — **Cloud SQL Studio reaches the private-IP instance with no infrastructure change**, so no Cloud Run Job was needed. KAN-217 (IAM database auth, deletion protection) was completed opportunistically in the same session and is **not** sprint scope.

| Scope | Result |
| --- | --- |
| Authenticated `(user_id, source_slug)` | **2 groups × 2 rows**, both `user_id = 1` |
| Guest `(guest_session_id, source_slug)` | **0 rows** |

**4 rows involved, 2 to delete.** Guest index needs no backfill → R3 costs nothing. D3's `single digits → S1 is a 1–2 day item` branch is the one taken, and that sized the box.

## A near-miss worth recording

Re-running the count without its two `WHERE` clauses reported `user_id 1 → 112`. **That is not 112 duplicates** — `GROUP BY user_id, source_slug` collapses every `NULL` into one bucket, so it counted *"user 1 has 112 recipes with no source"*. The same artifact made three guest rows look duplicated when they were different guests each saving one public recipe. The `WHERE` clauses are load-bearing.

Had that been taken at face value it would have flipped D3 to "the backfill *is* the sprint" and justified a content-hash schema change on a misread.

## S1 also carries a one-time public-surface purge

On Adam's direction — the live site has duplicates created before the guard exists. Test is **name AND image**; name alone gives candidates. There is **no image column**; image lives in the `data` JSON blob.

**KAN-194 changes the survivor rule here.** `togglePublic()` never backfills the image, so publish-toggle duplicates have **no image** — a strict same-name-AND-same-image match **won't find them**, and they're also the rows rendering a blank hero and blank OG. So for public rows: **keep the row with an image, delete the imageless twin**; fall back to oldest `created_at` only when both or neither have images; never delete `is_canonical`.

## Coverage limit, stated plainly

The indexes are **partial**. `NULL`-source rows — the majority — stay unconstrained. Correct by design (no provenance to collide on; a name-based constraint was rejected because two different recipes may share a title). The blind-spot probe found 4 same-name groups, only 2 confirmed same-user. **Same name is not the same recipe.** A content-hash column is a Sprint 7 candidate to be costed against evidence.

## New file

`specs/KAN-213_DEDUP_QUERIES.md` — every query, the survivor rule, and a **transaction-wrapped purge that ends in `ROLLBACK`** until deliberately changed.

---

https://claude.ai/code/session_012EfffNRnHWdCc7Pn4rEfMp

---

### DIFF
--- a/specs/KAN-213_DEDUP_QUERIES.md
+++ b/specs/KAN-213_DEDUP_QUERIES.md
@@ -3,8 +3,10 @@
 Working document for Sprint 6 / S1. Copy-paste into **Cloud SQL Studio**.
 Not synced to Confluence (same treatment as `specs/KAN-119_LOOP_PLAN.md`).
 
-> **Nothing here deletes anything until §5, and §5 runs inside a transaction that ends in
-> `ROLLBACK` until you deliberately change it.** Read a result before you act on it.
+> **Nothing here deletes anything until §5. Studio does not honour `BEGIN…ROLLBACK` blocks
+> reliably (§5 preamble), so §5 runs statements one at a time with a preview `SELECT`, an
+> `is_canonical = false` write guard, and a zero-row verify after each mutation.** Read every
+> result before running the next statement.
 
 ---
 
--- a/specs/KAN-213_DEDUP_QUERIES.md
+++ b/specs/KAN-213_DEDUP_QUERIES.md
@@ -101,6 +101,17 @@ FROM recipe WHERE is_public = true;
 
 There is **no image column** — image lives in the `data` JSON blob.
 
+> ### ⚠️ The image test does not work. Verified 2026-08-08.
+>
+> Every row's image resolves to **`/api/recipes/<its own id>/image`** — derived from the row id,
+> so **no two rows can ever match on image, duplicate or not.** §3c below is structurally
+> incapable of returning a row and is kept only as a record of the dead end. Comparing images
+> for real would mean hashing the GCS objects, which is not worth building for a one-time purge.
+>
+> **Use `source_slug IS NULL` instead** — it identifies the author's original, and everything
+> with a `source_slug` is a saved copy. That turned out to be a cleaner signal than image ever
+> would have been. See §3f.
+
 ### 3b. POTENTIAL duplicates — public name collisions, any owner
 
 ```sql
@@ -109,7 +120,7 @@ FROM recipe WHERE is_public = true
 GROUP BY name HAVING count(*) > 1 ORDER BY dupes DESC, name;
 ```
 
-### 3c. CONFIRMED duplicates — same name AND same image
+### 3c. ~~CONFIRMED duplicates — same name AND same image~~ — DEAD END, returns nothing
 
 ```sql
 SELECT name,
@@ -123,9 +134,11 @@ HAVING count(*) > 1 ORDER BY dupes DESC, name;
 ### 3d. The KAN-194 pattern — imageless twins
 
 `togglePublic()` never backfills the image from the source recipe, so publish-toggle
-duplicates have **no image**. Their pairs are (original _with_ image, duplicate _without_),
-so **3c will not match them** — they hide in 3b. They are also the rows rendering a blank
-hero and blank OG image on the public site.
+duplicates were expected to have **no image**, hiding from 3c and surfacing only in 3b.
+
+**Did not appear in the 2026-08-08 data** — all 11 public duplicate rows carry an image URL
+(self-referential, per the warning in 3a). Keep the query for future passes; it was not the
+mechanism behind this set.
 
 ```sql
 SELECT name, count(*) AS rows_,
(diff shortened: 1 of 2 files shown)


specs/KAN-213_DEDUP_QUERIES.md
```diff
@@ -0,0 +1,256 @@
...
+```sql
+SELECT name, count(*) AS dupes
+FROM recipe WHERE is_public = true
+GROUP BY name HAVING count(*) > 1 ORDER BY dupes DESC, name;
+```
+
+### 3c. CONFIRMED duplicates — same name AND same image
+
+```sql
+SELECT name,
```
adamtasteslikegood, line 128: **§3c "CONFIRMED duplicates" mislabels imageless name-collisions as confirmed.**

`GROUP BY name, COALESCE(ai_image_url, stock_image_url, image)` groups NULLs together (standard SQL `GROUP BY` semantics). Two *unrelated* public recipes that happen to share a name and both lack any image (all three JSON fields NULL) get bucketed here as "same name AND same image" — but they only share the *absence* of an image, not an actual matching one.

That matters because §4's survivor rule falls through to rule 3 (keep oldest `created_at`, delete the newer) for any pair where rule 2 doesn't apply — i.e. exactly this "both imageless" case. So a §3c "CONFIRMED" false positive can delete a genuinely distinct recipe, which is the exact failure mode the doc itself warns against in §6 ("same name is not the same recipe") but doesn't guard against here.

Suggest: filter `WHERE COALESCE(...) IS NOT NULL` in §3c so a NULL≡NULL match can't count as "confirmed," and route imageless name-collisions through §3d (which already exists for this) instead.

---

specs/KAN-213_DEDUP_QUERIES.md
```diff
@@ -0,0 +1,256 @@
...
+FROM recipe WHERE id IN ('<id-1>', '<id-2>');
+
+-- 2. refuse to proceed if any canonical row is in the list (must return 0)
+SELECT count(*) AS canonical_in_list
+FROM recipe WHERE id IN ('<id-1>', '<id-2>') AND is_canonical = true;
+
+-- 3. delete
+DELETE FROM recipe WHERE id IN ('<id-1>', '<id-2>') AND is_canonical = false;
+
+-- 4. re-run the §2 count — must now return 0 rows
```
adamtasteslikegood: **§5's postcondition check (step 4) only verifies the §2 invariant, but §5 is the shared template for both purges.**

The instructions above say "Replace the id list with ids you have actually read from §2b / §3e" — so §5 is meant to run for the public-surface purge (§3e ids) too. But step 4 always re-runs the `(user_id, source_slug)` count from §2, which has nothing to do with the public-surface duplicates §3e targets. If someone purges §3e ids, step 4 will report 0 rows regardless of whether the actual public-surface duplicate was removed (that count was likely already 0 going in) — giving false confidence that the purge worked when it verified an unrelated invariant.

Worth adding a second postcondition variant (re-run §3b/§3c grouped by `name`) for the §3e purge path, or noting explicitly that step 4 only validates the §2 branch.

---

adamtasteslikegood (commented):
Reviewed since Rovo Dev is out of credits on this one. This is documentation/runbook-only, but §5 of the new runbook is a template for real production `DELETE`s, so I read it as if it were code.

Two inline findings on `specs/KAN-213_DEDUP_QUERIES.md` worth fixing before this runbook is used against prod:
1. **§3c "CONFIRMED duplicates" can false-positive on two unrelated recipes that share a name and both lack an image** (SQL `GROUP BY` treats NULL≡NULL), which then falls through §4's survivor rule to delete a genuinely distinct recipe — the exact "same name ≠ same recipe" failure the doc warns about elsewhere but doesn't guard against here.
2. **§5's postcondition check (step 4) only validates the §2 `(user_id, source_slug)` invariant**, even though §5 doubles as the template for the §3e public-surface purge — running it with §3e ids won't actually verify the public duplicate was removed.

Also a handful of doc-consistency nits in `specs/SPRINT_6_PLAN.md` (not inline-commentable — they land outside this diff's hunks): the Gates table (~L358) still shows the duplicate count as blocked, charter row D5 (~L58) still calls R2 "most likely" while the R2 section (~L294) says "RETIRED," the "Blocker, open at charter time" paragraph (~L170) is still future-tense about an investigation that already happened, and the guest-duplicate SQL comment (~L153) still reads as an open TODO despite the resolved 0-row count just above it. None of these block anything functionally — just a few spots that weren't updated to match this PR's new "count is done" narrative, worth a quick consistency pass.

Everything else — the survivor rule, the canonical-row guard, the transaction-wrapped `ROLLBACK`-first purge — looks careful and safe.

---

adamtasteslikegood:
## Both outstanding findings addressed — commit `1008a3c`

### 1. `SPRINT_6_PLAN.md:195` — stale image-based survivor rule (**fixed**)

Verified before acting rather than taking the flag at face value: `SPRINT_6_PLAN.md` stated *"Keep the row with an image; delete the imageless twin"* as the current rule, while `KAN-213_DEDUP_QUERIES.md` §4 line 219 carries `~~Keep the one WITH an image~~ — dead, see §3a`, and §3a proves the criterion is structurally unusable (every image resolves to `/api/recipes/<its own id>/image`, so no two rows can ever match). The finding is correct, and it was worse than reported on one axis the three prior passes did not name: the blockquote also said **delete**, where §5a settled on **unpublish** for public rows.

Took the second of the two options offered — preserve the arc, remove the stale instruction. The blockquote is struck through with a line saying it was retired before it was ever applied and kept because the sizing rested on it; the replacement rule (`source_slug IS NULL` wins, canonical-copy exception, `created_at` fallback, never touch `is_canonical`) now appears directly beneath it, along with the delete-vs-unpublish correction and why.

### 2. §5c's missing preview `SELECT` (**moot — the statement is now refused outright**)

The flag was right that §5c diverged from the §5 preamble. It resolved in a direction worth recording: the pending statement is not getting a preview, it is **not being run at all**.

Its stated premise was that it *"prevents a dead 'source' link on `/r/vegan-cornbread` (KAN-212 class)"*. The code disproves that:

- `src/utils/public-link.ts:70` — `publicLinkKind()` returns `'own'` as its **first** branch when `is_public === true && slug`. Row `7e480d4b` is published under its own clean slug, so control never reaches the `sourceSlug` branch.
- `publishToggleKind()` is therefore `'normal'`, so the toggle title interpolating `sourceSlug` (`recipe-view.base.ts:325`) never renders either, and the KAN-137 confirm at `:252` is guarded by `!recipe.slug` so it does not fire.

Every path that reads `sourceSlug` is short-circuited. Nulling the column changes no rendered output — while destroying provenance and **removing the row from the new indexes**, which are both `WHERE source_slug IS NOT NULL`. The latent case (a dead link *if* the row is ever unpublished) is not a KAN-212 regression: `public-link.ts:14–16` and `:61–63` document it as a deliberate accepted trade-off, the guard being *"deliberately narrow"* on purpose.

### §5c is closed, and the gate is verified

Adam closed the Cornbread group through the app's publish toggle rather than SQL. The public-name-collision gate was verified **without database access**, from the surface the query describes — `/browse` renders every `is_public = true` row's `name` in the card `alt`, so the public index *is* the query's result set:

| Check | Result |
| --- | --- |
| Distinct public recipes across `/browse` | **76** |
| `/sitemap.xml` `/r/` entries | **76 — identical set**, zero diff |
| Public rows sharing a `name` | **0** ← the gate |
| `/r/vegasssdsdn-cornbread` | **404**, absent from sitemap ✅ |
| `/r/vegan-cornbread` | **200**, present in sitemap ✅ |

Two independently generated views of the same `is_public` set agreeing exactly is what makes the count trustworthy — a missed page would show as a diff.

Both purge gates now return zero rows, so **S1 moves to code**: the Alembic migration plus the concurrent-`POST` test, confirmed failing on today's code first.

One note for the record, since this PR has now produced four of them: the `source_slug = NULL` statement is the **same failure class as the R1 reversal** — a plausible-sounding claim about what the code does, written without reading the code, that would have caused a needless destructive `UPDATE`. Written up in runbook §5c rather than left in the commit log.

— _Replied by Claude on Adam's behalf_

